PR #9, opened 2026-04-17
Run dom-tester tests with a fixed pixel ratio

swankjesse:
Also never show scrollbars.

Ideally this makes the tests consistent on all the devices we use.


support/dom-tester/gradle-plugin/src/main/resources/com/wasmo/domtester/gradle/generated-dom-tester.js
```diff
@@ -0,0 +1,102 @@
...
+ */
+function configureBrowser(config) {
+  config.customLaunchers = config.customLaunchers || {};
+  config.customLaunchers['DomTesterChrome'] = {
+    base: 'ChromeHeadless',
+    flags: ['--screen-info={3200x3200 label=1st devicePixelRatio=2}'],
+    displayName: 'DomTesterChrome'
+  };
+  config.browsers = ['DomTesterChrome'];
+}
```
swankjesse, line 102: this is the only part that’s new

support/dom-tester/src/jsMain/kotlin/com/wasmo/domtester/Stylesheets.kt
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Customize the page to never measure the scrollbar gutter. Otherwise, it's derived from the host
+ * computer and therefore environment-dependent.
+ */
+val DomTesterStylesheet = """
+  * {
+    scrollbar-width: none;
+    scrollbar-gutter: stable both-edges;
+  }
```
swankjesse, line 17: This was difficult to figure out!

PR #14, opened 2026-04-28
fix ./gradlew os:server:server-development:build

15characterlimi:
distTar was pulling in two artifacts with the same archive path, `server-development/lib/real-jvm.jar`

This was because several modules end with `:real`, and their JVM jars are probably all being published with the same archive base name `real`.

This broke the build:

```
$ ./gradlew os:server:server-development:build
Execution failed for task ':os:server:server-development:distTar'.
> Entry server-development/lib/real-jvm.jar is a duplicate but no duplicate handling strategy has been set. Please refer to https://docs.gradle.org/9.2.1/dsl/org.gradle.api.tasks.Copy.html#org.gradle.api.tasks.Copy:duplicatesStrategy for details.
```

I'm fixing this by setting an archiveBaseName for each <Jar> task, based on the respective task's path. The gradle target path seems to always start with ":", so without the `removePrefix()` we'd get names like `-os-server-server-development.jar`.

Result:

```
$  tar -tf os/server/server-development/build/distributions/server-development.tar | grep '^server-development/lib/.*\.jar$'
server-development/lib/os-server-server-development.jar
server-development/lib/os-server-ktor-jvm.jar
server-development/lib/apps-journal-wasmo-app-jvm.jar
server-development/lib/apps-samples-jvm.jar
server-development/lib/os-server-accounts-real-jvm.jar
server-development/lib/os-server-computers-real-jvm.jar
server-development/lib/os-server-website-real-jvm.jar
server-development/lib/os-server-calls-real-jvm.jar
server-development/lib/os-server-installedapps-real-jvm.jar
[...]
```

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -3,6 +3,7 @@ import org.jetbrains.kotlin.gradle.dsl.KotlinJvmExtension
 import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
 import org.jetbrains.kotlin.gradle.plugin.KotlinMultiplatformPluginWrapper
 import org.jetbrains.kotlin.gradle.plugin.KotlinPluginWrapper
+import org.gradle.jvm.tasks.Jar
 
 plugins {
   id("wasmo-build").apply(false)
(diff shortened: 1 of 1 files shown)


15characterlimi:
Note: I'm not a gradle expert, it could be that there is a cleaner way to force unique names. TBH I'm surprised that this disambiguation doesn't happen by default.

15characterlimi:
P.S. Jesse mentioned that he only ever uses `:run` instead of `:build`. If y'all never run anything that depends on `distTar` or similar then perhaps we can just rip them out of the gradle configuration (I think they get added implicitly, probably by some of the standard Gradle plugins). But my guess is that sooner or later, we'll need them.

swankjesse (approved):
Love it

PR #18, opened 2026-05-03
Plan for local accounts

15characterlimi:
Jotting down my plan to align on the approach.

PR #19, opened 2026-05-03
[WIP draft] implement local accounts and auto sign-in

15characterlimi:
[WIP draft] implement local accounts and auto sign-in

This PR builds on top of https://github.com/wasmcomputercompany/wasmo/pull/18 - the first commit is from there.

Commits:

1. Write up a plan (local_accounts.md)
2. Small refactor to prepare for auto sign-in implementation
3. Implement local accounts and auto sign-in; doesn't quite work yet; AutoSignIn does the expected thing, but empirically the browser shows the same homepage as before, so perhaps the sign-in happens too late or it still needs to be tied to the cookie.
4. WIP change to safeguard email / passkey registration to fail if accountType = Local


swankjesse:
Okay, I’ve gone through the PR and I think I’d like for us to land some small ’foundational infrastructure’ PRs first, so that we can layer this behavior on top without as much effort.

Wanna help with these?

In particular, here’s some stuff that seems unduly difficult in this PR...

### Eager vs. On-demand Initialization

I don’t like that the AutoSignIn code has a side-effect of creating an account. I think instead we aught to create any default data that we’ll need when we start the service. Initially this is gonna be the admin username. Later we might have a privileges table (or whatever) that are associated to that, and I think that stuff all belongs in ‘initialize the app DB’ and not ‘get or create’. A pattern that worked well I’ve seen is something like this:

```kotlin
data class InitializationData(
  val adminAccountId: AccountId,
)

class OsInitializer(...) {
  suspend fun initialize(): InitializationData { ... }
}
```

Once we have that stuff, we can put `InitializationData` into the DI graph so anything that needs it can inject it.

As a bonus, it would be EXTREMELY RAD if the Homelab can do its own schema migration stuff on first boot. You already proposed this elsewhere and I think it’s all a part of the same system. The service boots up, it looks at its DB, and it gets the DB ready for service.

### Policy Enum vs. Separate DI Graphs

One stark thing we’re broken on now is there’s the same DI graph for all distributions. This is broken! The Homelab distribution shouldn’t know about Stripe for example.

But, I expect we should share about 90% of the DI graph between our distributions. 

Your PR calls attention to this with the `AccountType` enum, which exists to influence behavior downstream in the system. I’d prefer for the DI graph for each distribution to somehow have different implementations of the interfaces, rather than having `ConfirmEmailAddressRpc` detecting if it shouldn’t be called.

I’ll put some of this into motion ’cause I’m proposing something rather abstract here, but I think our code will be happier over the long term if we force this refactoring sooner than later.

swankjesse:
I also have feedback on the behavior changes here, but they’re not as spicy?!

I would prefer we add a UI for users to explicitly claim ‘I am admin’ vs. doing it automatically on the first action. The web app isn’t well suited to the implicit thing because our service goes out of its way to avoid creating account IDs for actions that don’t have side-effects.

Lastly, the most interesting part of of the `DbPassword` table is that it has a username. We don’t even assign a password! I think it’s a big insight that we need username+password auth for homelab, and I’m here for it. Would you consider changing the table name from `DbPassword` to `DbUsernameAndPassword` (or `DbLogin` if that name is too much) ? Heck, we could also do `DbUsername` only, and hold off on password until we implement OWASP best-practice password storage stuff.

15characterlimi:
Thanks for the care you put into your review comments. I'll adjust course. Some notes:

- I'd like to have distribution-specific source sets. I'll look into how to set this up with Gradle. Or perhaps Sandbox and Hosted should be the same flavor (more similar to each other than to Homelab). I think this will resolve several of the issues you observed, but we don't have it yet (AFAIK).

- Naming "Password" vs. "UsernameAndPassword" vs. "Username": This was a deliberate choice in  https://github.com/wasmcomputercompany/wasmo/pull/18  . My rationale was that the password part is the load bearing authn method difference - an absent password is arguably still an (empty) password, qualitative different from passkey / email, and a username is also "kind of" a password. I don't feel strongly, happy to change to UsernameAndPassword per your preference. **Edit** starting with Username SGTM.

- Agreed that admin account being constructed when the db is first initialized is cleaner. Perhaps this can be combined with distribution-specific sources, e.g. there could be a Homelab-specific .sql file that does the initialization.

PR #25, opened 2026-05-05
Work on account modeling

swankjesse:


docs/platform/plans/account_assets.md
```diff
@@ -0,0 +1,67 @@
...
+Passwords will be stored following [OWASP guidelines].
+
+### `InstanceAdminGrant`
+
+If an account has this asset, it has admin privileges on the instance.
+
+Admins may do the following:
+
+- Create invite links
+- Create ‘reset password’ links for any user
```
15characterlimi, line 34: Great job! FTR:

(a) This idea is better than 
(b) my original idea  of an antisymmetric, transitive (?) relation _isAdminFor_ between _pairs of accounts_. Initially, A is-admin-for B if A created account B.

Pro (a): 

- Matches root/admin accounts e.g. in Linux/Windows (familiarity)
- Generalizes better towards more general ACLs.
- Model for changing admin status follows simply

Pro (b): 
- Naturally expresses more hierarchy levels (school-wide admin, vs. a teacher is admin for their class). Perhaps unimportant for homelab.

Cons (b)
- Not clear how is-admin-for relation can dynamically update; do we need to avoid cycles? We don't want users to get stuck just because they weren't aware of the consequences of which order they created their accounts in.

So: ship it!

docs/platform/plans/account_assets.md
```diff
@@ -0,0 +1,67 @@
...
+redo that flow (but keep the passkeys).
+
+### `VerifiedEmailAddress`
+
+We [verify email addresses](../../code/email_address_linking.md).
+
+### `AcceptedTerms`
+
+We will add a terms of use document and require users to accept these terms before they can engage
+with the services we host.
```
15characterlimi, line 53: Optional: Perhaps mention: Acceptance of Terms (timestamp, version/exact text of terms, text of acceptance action) is also written to audit log. 

---

Perhaps we need a separate plan document on Audit Log. In my mind, Audit log entries consist of:
 - timestamp
 - principal [Session info, or system service], where Session info includes AccountId or logged-out session ID, originating IP address
 - payload

Deleting an account deletes the audit log entries for that account (FOREIGN KEY ON DELETE CASCADE?), but there's also an index to efficiently traverse by timestamp, and a snapshot of some _attributes_ that follow from the audit log'ed actions (e.g. isAdmin) are also stored elsewhere.

docs/platform/plans/account_assets.md
```diff
@@ -0,0 +1,67 @@
+Accounts Assets
+===============
+
+Different distributions require different assets before they are permitted to create or operate
+computers.
```
15characterlimi, line 5: Not convinced by the term "Account Assets" and the focus on "create or operate computers". 

The middle ground between (a) and (b) below that this doc's scope picks feels arbitrary:

(a) focus on "Signing in" (discussing how this differs per distribution, with a brief mention that signed-in accounts are allowed to do more), or
(b) general case: _User Actions_ require _Session_ to have particular _Attributes_. _User Actions_ is not just creating an account, _Session_ is not just signed-in, and _Attributes_ is not just signedInState, isAdmin, and ToS acceptance.

Would it be better to go all the way to either (a) or (b)? WDYT?

----

In case (b), the text up to here becomes:

Session Attributes
===============

Some actions require the current session to have particular attributes. The Action _create a computer_ requires attributes _isSignedIn_ and _lastAcceptedTermsOfServiceVersion >= N_ at that time. _resetAnotherUsersPassword_ only exists on homelab, and requires attribute _isAdmin_ (implies _isSignedIn_).

Ways to achieve `isSignedIn` differ per distribution (table below). On Sandbox and Hosted, _create a computer_ additionally requires  _accountBalance > 0_ [^1], but on homelab it does not (there's no _accountBalance_ concept on homelab).

[^1]: It might also require _userCountry not on sanctioned list_, but hopefully Wasmo can remain agnostic - leave this to our payments provider to enforce.

Attributes may take on two or more values, and possible value transitions may or may not be restricted.

Let's discuss some of these attributes in more detail, starting with signing in.

------

In case (a), the title would become "Signing in" and we'd write less.

WDYT?

15characterlimi (commented):
Top-in-document comment is load bearing.
Middle comment is praise + pondering
Bottom comment is a future separate document

Annoyingly, GitHub renders the comments in a different order below :)

docs/platform/plans/account_assets.md
swankjesse: Hell yes.

(Also audit logs are awesome! And we can do fancy things like sign records so it’s super difficult to be Dennis Nedry and turn the safety systems off)

docs/platform/plans/account_assets.md
swankjesse: Yeah, you’re right that my doc is both too abstract (assets) and too concrete (computer actions).

I’ve been reluctant to use the term ‘Signed In’ in the data model. I don’t like how it feels like a state (“signed in”) and not a computation (“you have met the requirements to perform common actions”). But I’m jumping through hoops to avoid it.

Lemme merge as-is, and promise to rework that whole part.

15characterlimi: I think state vs computation is an implementation detail. State is just a cached result of a computation: 

- _signed-in_ or _isAdmin_ can be computed from the event log, but it's more efficient to keep a snapshot of the value up to date, and read that
- _hasPositiveAccountBalance_ is computed ad-hoc from _accountBalance_, which is cached but can be recomputed if we discover a discrepancy and want to balance our books.

swankjesse: Update! https://github.com/wasmcomputercompany/wasmo/pull/27/changes

PR #26, opened 2026-05-06
Use runBlocking() instead of suspend fun main

swankjesse:
I think 'suspend fun main' is a trap, it doesn't have a good enough dispatcher to resume a continuation.

I suspect the underlying cause of this deficiency is that 'suspend' is language feature, but 'runBlocking' is in the coroutines library.

Either way, this fixes the app launch to use the main dispatcher before and after the database migration is performed.

### DIFF
--- a/os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/Postgresql.kt
+++ b/os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/Postgresql.kt
@@ -1,17 +1,33 @@
 package com.wasmo.sql
 
+import io.vertx.core.Completable
 import io.vertx.core.Future
 import io.vertx.sqlclient.Row
 import io.vertx.sqlclient.RowSet
 import io.vertx.sqlclient.SqlClient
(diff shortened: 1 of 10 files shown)


os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/Postgresql.kt
```diff
@@ -1,17 +1,33 @@
...
 suspend inline fun SqlClient.execute(
   sql: String,
 ): RowSet<Row> {
-  return query(sql).execute().asDeferred().await()
+  return query(sql).execute().awaitSuspending()
 }
 
-fun <T> Future<T>.asDeferred(): Deferred<T> =
-  toCompletionStage().asDeferred()
+/** @param onCancellation to release the result value if it holds any resources. */
```
swankjesse, line 18: Not strictly necessary, but `suspendCancellableCoroutine` has fewer indirections vs. `asDeferred().await()`, and as an important bonus we can avoid leaking a connection when the coroutine is canceled while waiting

PR #29, opened 2026-05-07
Introducing distribution codes and using it for app databases/users.

mpawliszyn:


os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
```diff
@@ -0,0 +1,9 @@
+package com.wasmo.identifiers
+
+enum class DistributionCode(val code: String) {
```
15characterlimi: Is there a reason you have this enum in addition to identifiers/src/commonMain/kotlin/com/wasmo/identifiers/DistributionCode.kt ? Looks like os/server/identifiers depends on implementation(projects.identifiers)

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
```diff
@@ -0,0 +1,9 @@
+package com.wasmo.identifiers
+
+enum class DistributionCode(val code: String) {
```
15characterlimi: I guess the reason you made this enum rather than keeping each of these codes inside the distributions that use them (e.g. HomelabWasmoOs knowing about "hl") is because you wanted to make it clear that they're different? Okay.

Brainstorming: Given that the enum does exist and lives in identifiers, how about DistributionIdentifiers.shortCode ?

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
```diff
@@ -0,0 +1,9 @@
...
+  HomeLab(code = "hl"), // localhost
+  Hosted(code = "wc"),  // wasmo.com
+  Sandbox(code = "wd"), // wasmo.dev
+  SDK(code = "sd"),
```
15characterlimi: nit: I come from a school where acronyms are camelCased/PascalCased (HttpUrl, not HTTPURL). So this should probably be Sdk?

15characterlimi (commented):
Hi! Feel free to ignore my unqualified comments, I'm new here :-)

mpawliszyn:
> Hi! Feel free to ignore my unqualified comments, I'm new here :-)

Your comments are greatly appreciated!

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
mpawliszyn: Yeah this is my mistake. I did a move without a proper intellij move... oops

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
mpawliszyn: Yep great point

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
mpawliszyn: I'm not sure I fully follow what you are proposing here. I agree that shortCode is better.

What is the container for DistributionIdentifiers. Although maybe you are referring to my mistake of keeping the copy in identifiers.

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
mpawliszyn: Totally missed that I kept the original copy too. 🤦 

Thanks for the catch!

mpawliszyn:
I like having the enum value but with the shortcode so that we get type safety and an obvious reference to the deployment type but also with an obvious mapping to the shortcode.

swankjesse:
Maybe do something like `ComputerSlug`, where there’s a `DistributionCode` that just wraps a string? That thing could be in the main DI container.

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/DistributionCode.kt
15characterlimi: The alternative I was alluding to was: Instead all the shortCodes living here in this enum, they could be distributed across {Homelab,Sandbox,Hosted}WasmOs.kt.

Advantage:
- Distribution-specific stuff (e.g. Homelab's shortCode "hl") lives in distribution-specific source files and doesn't even exist in the binary for e.g. Sandbox

Downside:
- It's harder to verify if the shortCodes for the different distributions are distinct (we could have a unit test enforcing this?).

Probably overall it's not worth it, so your approach is fine. Just wanted to make sure we make a considered decision.

-----

On naming: `DistributionIdentifiers.shortCode` would mean renaming the enum `DistributionCode` -> `DistributionIdentifiers`, and renaming the enum attribute `code` -> `shortCode`.

Perhaps DistributionSlug.value would also work. This project is the first time I heard the term "slug", so I don't know if it fits. Use your judgement

swankjesse: I prefer DistributionCode as a value object.

Slug is a term I’ve seen to describe parts of URLs. And that’s what we’re using ’em for.
https://stackoverflow.com/a/4230937/40013

mpawliszyn: Maybe this is better this way since we will generally be trusting the value verbaitim so I'm fine with a value too.

I still think having a reference that is easy to discover, nearer to the code than just the readme, is likely important but very open to where that is.

mpawliszyn:
I went with DistributionShortCode and changed it to a value. I'm not in a hurry to merge so I'll wait until monday in case folks have anything else to add.

swankjesse (approved):

PR #30, opened 2026-05-07
Username table + admin account

15characterlimi:
- Introduce a Username table (DB schema revision 4)
- In Homelab only, as part of upgrade to revision 4, add a account with username "admin".
  - This is done by refactoring `Migrate` into a class with distribution-specific instances. The implementation takes optional  `customPostUpgradeSteps`, which Homelab uses to create the acount.
  - In future, different distributions could have more radically different databases. That's why `Migrate`, not just `customPostUpgradeSteps`, is now customizable per Distribution. E.g. we might not even create the Username table outside Homelab.
- Please review whether (a) the Dependency Injection setup, and (b) the `customPostUpgradeSteps` being bound in HomelabGraph, make architectural sense.


docs/local_development/absurd.md
```diff
@@ -5,7 +5,10 @@ Absurd powers some workflow code in Wasmo OS.
 
 ℹ️ Prerequisite: [PostgreSQL](postgresql.md).
 
-Only install it if you’d like to use their tooling to manage workflows.
+It's automatically installed in the Wasmo DB as part of the DB schema migrations (`ensureSchemaVersion()`).
```
swankjesse, line 8: NICE

docs/platform/plans/db_schema_migrations.md
```diff
@@ -34,6 +34,22 @@ ON CONFLICT (id) DO NOTHING;
...
+
+absurd is installed as part of the Wasmo DB as part of the automatic schema migrations.
+
+ - Wasmo schema DB version 2 includes absurd schema version 0.3.0 (like `absurdctl init --ref 0.3.0`)
+ - Wasmo schema DB version 3 includes an absurd queue named `default` (like `absurdctl create-queue default`)
+
+For future upgrades, we'll have to generate Wasmo DB schema migration SQL through a command like
+the below, but that requires our PRs to land in absurd, first.
+
+```bash
```
swankjesse, line 47: this is quite excellent

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabGraph.kt
```diff
@@ -58,12 +62,17 @@ import wasmo.sql.SqlDatabase
...
+internal interface HomelabGraph : DistributionGraph {
+  override val wasmoService: WasmoService
+  override val migrate: Migrate
   val callGraphFactory: CallGraph.Factory
   val computerServiceGraphFactory: ComputerServiceGraph.Factory
   val installedAppServiceGraphFactory: InstalledAppServiceGraph.Factory
 
+  @Provides
+  @SingleIn(OsScope::class)
+  fun provideMigrate(clock: Clock): Migrate = HomelabMigrateImpl(clock)
```
swankjesse: I like these to have operator names like `Migrator`

swankjesse: And `Impl` as a suffix is forbidden, you gotta use `Real` as a prefix instead!

swankjesse: And finally, you should be able to do get Metro to call your constructor for you automatically

```
@Binds
abstract fun bindMigrate(real: RealHomelabMigrate): Migrate
```

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabMigrateImpl.kt
```diff
@@ -0,0 +1,20 @@
+package com.wasmo.distributions.homelab
+
+import com.wasmo.db.MigrateImpl
+import com.wasmo.db.NamedSchema
+import com.wasmo.db.accounts.insertAccount
+import com.wasmo.db.usernames.insertUsername
+import kotlin.time.Clock
+
+internal class HomelabMigrateImpl(clock: Clock) : MigrateImpl(
```
swankjesse:
```
@Inject
@SingleIn(OsScope::class)
```

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/RealHomelabMigrator.kt
```diff
@@ -0,0 +1,20 @@
+package com.wasmo.distributions.homelab
+
+import com.wasmo.db.MigrateImpl
+import com.wasmo.db.NamedSchema
+import com.wasmo.db.accounts.insertAccount
+import com.wasmo.db.usernames.insertUsername
+import kotlin.time.Clock
+
+internal class HomelabMigrateImpl(clock: Clock) : MigrateImpl(
+  customPostUpgradeSteps = mapOf(
```
swankjesse, line 16: I think this `customUpgradeSteps` could just be a function?


```
fun upgradeDistribution(...) { ... }
```

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabWasmoOs.kt
```diff
@@ -43,11 +43,11 @@ fun main(args: Array<String>): Unit = runBlocking {
     override val provisioningPostgresqlAddress: PostgresqlAddress
       get() = sharedPostgresqlAddress
 
-    override fun createService(
+    override fun createServiceGraph(
```
swankjesse: NICE

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabGraph.kt
```diff
@@ -58,12 +62,17 @@ import wasmo.sql.SqlDatabase
     WebsiteBindings::class,
   ],
 )
-internal interface HomelabGraph {
-  val wasmoService: WasmoService
+internal interface HomelabGraph : DistributionGraph {
```
swankjesse: I don’t think I like `DistributionGraph` as an interface; instead put `migrate` as a member of `WasmoService`.

The graph produces the `WasmoService`, which is the shared thing that all distributions have

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/DbUsername.kt
```diff
@@ -0,0 +1,12 @@
...
+
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.UsernameId
+import kotlin.time.Instant
+
+data class DbUsername(
+  val id: UsernameId,
+  val createdAt: Instant,
+  val accountId: AccountId,
+  val username: String,
```
swankjesse: We should probably set up something like `UsernameSlug` for username, perhaps just call it `Username`? That way we can lock down what it contains and get a bit of type safety to boot

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
```diff
@@ -8,53 +8,84 @@ import okio.Path.Companion.toPath
...
-    val match = MigrationNameRegex.matchEntire(path.name)
-      ?: error("unexpected migrations path: $path")
-    Migration(match.groupValues[1].toInt(), path)
-  }
+interface Migrate {
+
+  /**
+   * Upgrades the DB schema to the given version, if needed.
+   *
+   * @param targetSchema the schema version to upgrade to, or null to upgrade to the latest version.
```
swankjesse, line 19: Why would we ever migrate to not the latest version?

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrate.kt
```diff
@@ -8,53 +8,84 @@ import okio.Path.Companion.toPath
...
-    val migrationSql = FileSystem.RESOURCES.read(migration.path) {
-      readUtf8()
+    val migrations = migrationPaths.map { path ->
+      val match = MigrationNameRegex.matchEntire(path.name)
+        ?: error("unexpected migrations path: $path")
+      Migration(match.groupValues[1].toLong(), path)
+    }.groupBy { it.version }
+
+    for (version in (oldVersion + 1)..newVersion) {
+      migrations[version] ?.forEach { migration ->
```
swankjesse: supernit: I also ban (BAN!) `forEach` because it doesn’t step-debug as well as `for(...)`

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
```diff
@@ -8,53 +8,84 @@ import okio.Path.Companion.toPath
...
 private data class Migration(
-  val version: Int,
+  val version: Long,
   val path: Path,
 )
+
+private val CURRENT_SCHEMA_VERSION = NamedSchema.entries.maxBy { it.version }
+
+/**
+ * Enum instances to make schema revision versions more readable.
```
swankjesse, line 88: Supernit: I would prefer to not name the schema versions. In my last project we had hundreds of 'em and the names aren’t that interesting?

swankjesse: (the migration files have names, and that’s enough?)

os/server/db/src/jvmMain/resources/migrations/v2__absurdctl_init_ref_0.3.0.sql
```diff
@@ -0,0 +1,1685 @@
+-- Absurd installs a Postgres-native durable workflow system that can be dropped
```
swankjesse, line 1: This is available at `/Volumes/Development/wasmo/support/absurd/vendor/absurd/sql/absurd.sql`

os/server/db/src/jvmMain/resources/migrations/v3__absurdctrl_create-queue_default.sql
```diff
@@ -0,0 +1 @@
+ SELECT absurd.create_queue('default');
```
swankjesse, line 1: excellent

os/server/db/src/jvmMain/resources/migrations/v2__absurdctl_init_ref_0.3.0.sql
swankjesse: And that’s okay. We can have two of 'em, and delete the other one when we promote it upstream)

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE Username (
+  id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  created_at TIMESTAMPTZ NOT NULL,
+  -- 0 or 1 usernames per account
```
swankjesse: (This requirement likely won’t last forever. Changing your username could be a thing, as could be holding old usernames or redirecting 'em.)

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
```diff
@@ -10,18 +10,23 @@ import io.ktor.server.netty.EngineMain
 import wasmo.sql.SqlDatabase
 import wasmox.sql.transaction
 
+interface DistributionGraph {
```
swankjesse: As above, please remove?

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
```diff
@@ -34,17 +39,17 @@ abstract class Distribution {
...
       provisioningDb = provisioningDb,
       wasmoDb = wasmoDb,
     )
 
+    wasmoDb.transaction {
+      distributionServiceGraph.migrate.ensureSchemaVersion()
+    }
+
+    val wasmoService = distributionServiceGraph.wasmoService
     wasmoService.start()
```
swankjesse, line 41: Should `WasmoService.start()` do this start-the-service stuff?

swankjesse (approved):
Looks excellent. Don’t mind my abundance of comments, it’s almost all nitpicks.

docs/platform/plans/db_schema_migrations.md
15characterlimi: thanks for always being so gracious, hopefully soon I'll actually deserve it :-)

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
15characterlimi: In a unit test that confirms that a particular range of upgrade/migration steps has a particular result?

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/RealHomelabMigrator.kt
15characterlimi: Where would the function live? Note that the implementation is specific to a distribution.

Note right now the customPostUpgradeSteps are just the part that isn't shared between distributions and/or that calls Kotlin functions. I imagine the migration paths for the different distributions will look more radically different in the future.

The way it's written for now, each element here only has to care about one particular revision, mapping this to a range of revisions is handled by Migrate.

os/server/db/src/jvmMain/resources/migrations/v2__absurdctl_init_ref_0.3.0.sql
15characterlimi: As noted on the description of the base PR https://github.com/wasmcomputercompany/wasmo/pull/28 , I had copied it from there. Will look into a PR later to remove the original.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabWasmoOs.kt
15characterlimi: Reverting back to createService() in order to not need the HomelabGraph base interface

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
15characterlimi: Leaving as-is for now, adding this sentence to the comment: 

 If this ever changes, need to drop 'UNIQUE' here and on the index.

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
15characterlimi: Yeah makes sense for wasmoService.start() to do the ensureSchemaVersion(), it just means that:

1. WasmoService needs migrator and wasmoDb injected (can probably be composed into a WasmoDb class in future)
2. wasmoService.start() needs to become a suspend fun

Both seem reasonable, so I'm doing it.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/RealHomelabMigrator.kt
15characterlimi: I suggest I leave this as-is for now. In a future PR, I wanna unify the SQL- and Kotlin-driven schema upgrade steps, I think it'll become cleaner then.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabGraph.kt
15characterlimi: all done in this push (sorry I forgot to push an "address review comments" commit, but also this affected two different commits): https://github.com/wasmcomputercompany/wasmo/compare/871212a57cefd86fbb0f6c501162ac3128a3722b..fc4288f3bb445b5ff68d6b5cfd016b13c7a6aca7

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/DbUsername.kt
15characterlimi: Yes! String is the wrong type for username. For example, String == / hashCode may not be appropriate for username (username should probably be case insensitive?), and we probably want to forbid certain characters to make the username URL safe.

Not sure about Username vs UsernameSlug, but I'll go with UsernameSlug for now and will enforce a-z0-9._- or so in the ctor (lowercase only, so we need to normalize at username entry time).

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
15characterlimi: It's referenced from the customPostUpgradeStep in RealHomelabMigrator. I want that to say `NamedSchema.USERNAME.version` rather than `4`.

But this will change in a follow-up PR where I unify the SQL- vs. Kotlin-based parts of the upgrade steps. Stay tuned.

15characterlimi:
Heads up, I plan to submit this PR in ~ 12 hours unless you protest. If you wanna see what I changed in response to review comments, see: https://github.com/wasmcomputercompany/wasmo/compare/871212a57cefd86fbb0f6c501162ac3128a3722b..45e598fe612291dc123625ae95279a9e6297ffbd

Planned work for the next days:

 - unify SQL- and Kotlin-based DB migration steps
 - do more of the wiring for Username based accounts for Homelab
   - e.g. HomeUI and RealCallDataService need to change; TBD how

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/RealHomelabMigrator.kt
swankjesse: Works for me.

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
swankjesse: Sounds good.

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE Username (
+  id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  created_at TIMESTAMPTZ NOT NULL,
+  -- 0 or 1 usernames per account. If this ever changes, need to drop 'UNIQUE' here and on the index.
```
swankjesse, line 4: I think my preference would be to add something like `replaced_at` as a timestamp, and configure the unique index with `WHERE replaced_at IS NOT NULL`. That way you could remove a username, or change usernames, and still keep a record of who had a given username at what time

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/WasmoService.kt
```diff
@@ -2,23 +2,33 @@
...
 class WasmoService(
   private val server: EmbeddedServer<*, *>,
   private val actionRouter: ActionRouter,
   private val jobProcessor: JobProcessor,
+  private val migrator: Migrator,
+  private val wasmoDb: SqlDatabase,
 ) {
-  fun start() {
+  suspend fun start() {
+    // AbsurdOsJobProcessor from jobProcessor.start() depends on this already having run.
```
swankjesse, line 25: excellent

swankjesse (approved):

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
15characterlimi: Some thoughts:

- `replaced_at` duplicates the value `created_at` of the new row. How about a nullable `canonical_username`  or `canonical_id` or similar that points to the latest row for the same account? `WHERE account_id=42 ORDER BY create_at ASC` still gives us the order of replacements. `ORDER BY id` probably does. too.
- If we want to handle removal-without-replacement here in the `Username` table, how about `tombstoned_at` instead of `replaced_at`? Or if a username can only be deleted by closing the account, `tombstoned_at` could live on the `Account` table.

WDYT?

I've posted a message on the chat about how to preserve a clean migration history in the face of such schema fixes.

swankjesse: Yeah `tombstoned_at` or `deleted_at` is probably the semantics I want.

PR #32, opened 2026-05-09
Start to provide Distribution entrypoints other than for starting the WasmoOs server

15characterlimi:
This PR consists of multiple commits, I suggest you review them as individual steps.

Historically, each Distribution only does one thing (start WasmoService); entrypoint and DI graph are purely in service of that one purpose.

This PR proposes that additional entrypoints/tools that share most of the DI graph could be useful. The proof of concept is an entrypoint that exits just short of starting the WasmoService, but after having done the DB migration. This allows us to squash the DB migration steps (including absurd) into a single one, as follows:

 1. Drop and recreate the database (`psql` commands)
 2. run  `Distribution.prestartWasmoServiceAndExit()`
 3. Dump the database (`pgdump`).

**Jesse**: I know you didn't like additional purposes of the Graph but I kind of do. I'm curious what you think after this PR.

----

How:

To separate `WasmoService.start()` from pre-start activities,
 - DB Migration moves out of WasmoServer into Distribution
 - `Distribution.createService()` -> `createServiceGraph`. which requires the *Graph interfaces to derive from a parent interface `DistributionGraph`
 - `Migrator` moves up into `DistributionGraph` to become a sibling of `wasmoService`.

The new `HomelabPrestartWasmoServiceAndExit`:
 -  exposes `Distribution.prestartWasmoServiceAndExit()` as an entrypoint (`main` method) on Homelab. I see no reason not to do the same on Hosted and Sandbox.
 - Reuses `HomelabDistribution`, which moved from `HomelabWasmoOs.kt` into its own .kt file.

Future steps (copied from the last commit's message):
   1. Figure out how to expose HomelabPrestartWasmoServiceAndExit as a Gradle task.
   2. Drop the dependency from Distribution.commonInit() onto EngineMain.createServer(), because in commonInit we don't know yet if the server will actually be started. The complication is that server is then not available to be passed into createServiceGraph(), and there are at least two DI bindings that currently build onto server:
       - KtorLogger constructs `log = server.application.log`
       - ServiceBindings.provideApplication() constructs `application = server.application`
       - If we can construct DI bindings for both without going via server, it should be possible to pull server out of the DistributionGraph (construct it manually, or construct a separate graph for the server-specific parts of the system).

### DIFF
--- a/os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
+++ b/os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
@@ -28,7 +28,7 @@ abstract class Distribution {
     wasmoDb: SqlDatabase,
   ): DistributionGraph
 
-  suspend fun start(args: Array<String>) {
+  suspend fun startWasmoService(args: Array<String>) {
     val server = EngineMain.createServer(args)
 
     val postgresqlClientFactory = PostgresqlClient.Factory()
--- a/os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
+++ b/os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
@@ -1,5 +1,6 @@
 package com.wasmo.wiring
 
+import com.wasmo.db.Migrator
 import com.wasmo.sql.PostgresqlAddress
 import com.wasmo.sql.PostgresqlClient
 import com.wasmo.sql.ProvisioningDb
@@ -7,6 +8,12 @@ import com.wasmo.sql.asSqlDatabase
 import io.ktor.server.engine.EmbeddedServer
 import io.ktor.server.netty.EngineMain
 import wasmo.sql.SqlDatabase
+import wasmox.sql.transaction
+
+interface DistributionGraph {
+  val wasmoService: WasmoService
+  val migrator: Migrator
+}
 
 /**
  * Each subclass is its own particular distribution of Wasmo OS.
@@ -15,11 +22,11 @@ abstract class Distribution {
   protected abstract val osPostgresqlAddress: PostgresqlAddress
   protected abstract val provisioningPostgresqlAddress: PostgresqlAddress
 
-  protected abstract fun createService(
+  protected abstract fun createServiceGraph(
     server: EmbeddedServer<*, *>,
     provisioningDb: ProvisioningDb,
     wasmoDb: SqlDatabase,
-  ): WasmoService
+  ): DistributionGraph
 
   suspend fun start(args: Array<String>) {
     val server = EngineMain.createServer(args)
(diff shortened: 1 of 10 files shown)


os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
```diff
@@ -1,12 +1,19 @@
...
 import com.wasmo.sql.ProvisioningDb
 import com.wasmo.sql.asSqlDatabase
 import io.ktor.server.engine.EmbeddedServer
 import io.ktor.server.netty.EngineMain
 import wasmo.sql.SqlDatabase
+import wasmox.sql.transaction
+
+interface DistributionGraph {
+  val wasmoService: WasmoService
+  val migrator: Migrator
```
swankjesse, line 15: I would have expected that `WasmoService` would itself inject the `Migrator`

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
```diff
@@ -32,12 +43,36 @@ abstract class Distribution {
...
+  }
+
+  suspend fun startWasmoService(args: Array<String>) {
+    val wasmoServiceStarter = commonInit(args)
+    // start, never close()
+    wasmoServiceStarter.start()
+  }
 
-    wasmoService.start()
+  suspend fun prestartWasmoServiceAndExit(args: Array<String>) {
```
swankjesse, line 73: What’s this for?

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
15characterlimi: That was the case before this PR. It changes in this PR because we want to be able to run the Migrator without necessarily even constructing the Service. 

The Distribution now represents anything configured with DI bindings to do things compatibly with the service, but it doesn't necessarily run the service.

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
15characterlimi: To do everything that starting the service would do (up to and including the DB migration), but then exit short of actually starting the service.

Use case: We can inspect the DB immediately prior to the service starting. This allows us to dump the already-migrated DB contents at the latest schema revision, before the service has a change to modify the DB contents. This allows us to flatten the DB migrations into a single step. This mean we can rewrite DB migration history to lose all the messy in-between states _without_ losing data (for those who have already migrated to the latest version): we first write a migration that incrementally migrates data, then much later we collapse the migration history into a single step.

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
swankjesse: I don’t want a `Graph` interface. I think `WasmoService` should be the only type that the different distributions each produce.

I hear that you’re thinking we could offer other behaviours other than ‘run a web server’ but I think those things could be functions on WasmoService.

os/server/wiring/src/jvmMain/kotlin/com/wasmo/wiring/Distribution.kt
swankjesse: I don’t think we need this to be a part of the API though?

PR #34, opened 2026-05-10
start refactoring SignedIn state out of AccountSnapshot

15characterlimi:
First steps towards implementing [docs/code/accounts_and_authentication.md](https://github.com/wasmcomputercompany/wasmo/blob/8db5af7f3b821b4114aed7eeb914b7d310cea0cd/docs/code/accounts_and_authentication.md): signed-in status can be conferred by username [/password], not necessarily email addresses / passkeys.

- HomeUi only depends on `isSignedIn`
- Some tests depend on specifics
  - Factored out sealed `SignedInSnapshot` with implementations `EmailPasskeySignedInSnapshot` and `UsernameSignedInSnapshot` (unused)
  - Construction currently still happens inside `AccountSnapshot`; next step will be to construct server-side and change the `AccountSnapshot` wire format.

This PR consists of separate commits, in case you prefer to review them individually.

### DIFF
--- a/os/api/src/commonMain/kotlin/com/wasmo/api/AccountSnapshot.kt
+++ b/os/api/src/commonMain/kotlin/com/wasmo/api/AccountSnapshot.kt
@@ -9,13 +9,21 @@ import okio.ByteString
 @Serializable
 data class AccountSnapshot(
   val nextChallenge: @Serializable(Base64UrlSerializer::class) ByteString,
-  val passkeys: List<PasskeySnapshot>,
-  val emailAddresses: List<LinkedEmailAddressSnapshot>,
+  private val passkeys: List<PasskeySnapshot>,
+  private val emailAddresses: List<LinkedEmailAddressSnapshot>,
   val hasInvite: Boolean,
 ) {
+  // TODO: make signedInSnapshot a ctor argument (changes wire format), and support
+  // Username-based SignedInSnapshot server-side.
+  val signedInSnapshot: SignedInSnapshot? =
+    // TODO: We probably want "passkeys.isEmpty() &&", but that'll be a behavior change to before
+    if (emailAddresses.isEmpty()) {
+      null
+    } else {
+      EmailPasskeySignedInSnapshot(passkeys, emailAddresses)
+    }
   val isSignedIn: Boolean
-    // TODO: Do we also need "passkeys.isNotEmpty() ||" ?
-    get() = emailAddresses.isNotEmpty()
+    get() = signedInSnapshot != null
 }
 
 @Serializable
--- a/os/api/src/commonMain/kotlin/com/wasmo/api/AccountSnapshot.kt
+++ b/os/api/src/commonMain/kotlin/com/wasmo/api/AccountSnapshot.kt
@@ -12,7 +12,11 @@ data class AccountSnapshot(
   val passkeys: List<PasskeySnapshot>,
   val emailAddresses: List<LinkedEmailAddressSnapshot>,
   val hasInvite: Boolean,
-)
+) {
+  val isSignedIn: Boolean
+    // TODO: Do we also need "passkeys.isNotEmpty() ||" ?
+    get() = emailAddresses.isNotEmpty()
+}
 
 @Serializable
 data object AccountSnapshotRequest
--- a/os/api/src/commonMain/kotlin/com/wasmo/api/SignedInSnapshot.kt
+++ b/os/api/src/commonMain/kotlin/com/wasmo/api/SignedInSnapshot.kt
@@ -1,0 +1,24 @@
+package com.wasmo.api
+
+import com.wasmo.identifiers.UsernameSlug
+import kotlinx.serialization.Serializable
+
+@Serializable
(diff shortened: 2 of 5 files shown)


os/server/emails/real/src/jvmTest/kotlin/com/wasmo/emails/SignInSignOutTest.kt
```diff
@@ -21,46 +21,42 @@ class SignInSignOutTest {
     val client = tester.newClient()
 
     val confirmResponse = client.linkAndConfirmEmailAddress("jesse@example.com")
-    assertThat(confirmResponse.body.account?.emailAddresses)
+    assertThat(confirmResponse.body.account?.isSignedIn)
```
15characterlimi, line 24: we could potentially instead say

`.signedInSnapshot as? EmailPasskeySignedInSnapshot)?.emailAddresses`

and keep the isNotEmpty() below. Let me know what you prefer. If we go with isNotEmpty() then I'd prefer to still have one test that also checks that asserts that this translates into isSignedIn

os/api/src/commonMain/kotlin/com/wasmo/api/AccountSnapshot.kt
```diff
@@ -9,10 +9,22 @@ import okio.ByteString
 @Serializable
 data class AccountSnapshot(
   val nextChallenge: @Serializable(Base64UrlSerializer::class) ByteString,
-  val passkeys: List<PasskeySnapshot>,
-  val emailAddresses: List<LinkedEmailAddressSnapshot>,
+  private val passkeys: List<PasskeySnapshot>,
```
swankjesse, line 12: My initial intention with this thing (and everything in api) is to be a struct shared between client and server. Encapsulating state feels awkward for such data objects. (As would encapsulating state on a proto)

os/api/src/commonMain/kotlin/com/wasmo/api/AccountSnapshot.kt
```diff
@@ -9,10 +9,22 @@ import okio.ByteString
...
+  private val emailAddresses: List<LinkedEmailAddressSnapshot>,
   val hasInvite: Boolean,
-)
+) {
+  // TODO: make signedInSnapshot a ctor argument (changes wire format), and support
+  // Username-based SignedInSnapshot server-side.
+  val signedInSnapshot: SignedInSnapshot? =
```
swankjesse, line 18: I believe if we’re gonna do a SignedInSnapshot, then AccountSnapshot should be a member of that, instead of vice-versa.

(The session has exactly one account, but each account may have multiple sessions)

os/api/src/commonMain/kotlin/com/wasmo/api/SignedInSnapshot.kt
```diff
@@ -0,0 +1,24 @@
+package com.wasmo.api
+
+import com.wasmo.identifiers.UsernameSlug
+import kotlinx.serialization.Serializable
+
+@Serializable
+sealed class SignedInSnapshot
+
+@Serializable
+class EmailPasskeySignedInSnapshot(
```
swankjesse, line 10: This doesn’t track my expectation for how our accounts should work. In particular it’s not _either_ email or username, it’s at least one.

swankjesse: (The situation I’m imagining is hooking up a Homelab to be able to send email)

os/server/emails/real/src/jvmTest/kotlin/com/wasmo/emails/LinkEmailAddressTest.kt
```diff
@@ -36,7 +37,7 @@ class LinkEmailAddressTest {
       challengeCode = email.extractChallengeCode(),
     )
     assertThat(confirmResponse.body.decision).isEqualTo(Decision.LinkedNew)
-    assertThat(confirmResponse.body.account?.emailAddresses)
+    assertThat((confirmResponse.body.account?.signedInSnapshot as? EmailPasskeySignedInSnapshot)?.emailAddresses)
```
swankjesse, line 40: I don’t think we should need to downcast here, instead the signed in session should have an email address that might be null if the distribution doesn’t use that

os/server/emails/real/src/jvmTest/kotlin/com/wasmo/emails/SignInSignOutTest.kt
swankjesse: The big insight you made, that I am fully on board with, is that there’s a property on a session called 'isSignedIn', that’s determined by the server. I would like for us to add that to AccountSnapshot, and then perhaps split AccountSnapshot into two things, SessionSnapshot (identified by the cookie) and AccountSnapshot (identified by the accountid)

swankjesse (changes requested):

15characterlimi:
Thanks for your review! Based on your review comments, I'm closing this PR in favor of https://github.com/wasmcomputercompany/wasmo/pull/39

PR #35, opened 2026-05-12
More UI around signing up

swankjesse:


os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
```diff
@@ -23,8 +24,9 @@ class SignUpPresenter(
 ) : Presenter<SignUpModel, SignUpEvent> {
   private val mutableModel = MutableStateFlow(
     SignUpModel(
-      emailAddressCaption = "We’ll email you a challenge code",
-      challengeCodeCaption = "",
+      enterEmailAddressModel = EnterEmailAddressModel(
```
swankjesse: This is where we’re forcing everything into the Hosted, email-centric world. Next step is to provide the client with enough information to populate the `SelectUsernameToSignInModel` instead of this

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/api/RealWasmoApi.kt
```diff
@@ -61,6 +65,18 @@ class RealWasmoApi : WasmoApi {
     return call("/create-computer-spec", request)
   }
 
+  override suspend fun createUsername(
+    request: CreateUsernameRequest,
+  ): CreateUsernameResponse {
+    return call("/create-username", request)
```
swankjesse, line 71: These APIs don’t exist on the server yet!

os/api/src/commonMain/kotlin/com/wasmo/api/Username.kt
```diff
@@ -0,0 +1,19 @@
+package com.wasmo.api
+
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class CreateUsernameRequest(
+  val username: String,
```
15characterlimi: is there a reason why you use `String` for usernames in some places, rather than `UsernameSlug`?

In my mind, only URL safe `UsernameSlug` values are permissible usernames. Even UsernameQueries uses UsernameSlug.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/api/RealWasmoApi.kt
15characterlimi: Do I understand it right that /create-username is for creating a username *without* signing in? E.g. `admin` will create the initial additional usernames, but those are for other people. Whereas linkUsername is to sign-in and tie the username to the current account. Right?

Side note: In the distant future, we probably also need a dry-run version of this endpoint so that the client can check if a username is available *before* the user clicks submit (which then fails)

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SelectUsernameToSignInScreen.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.jetbrains.compose.web.dom.Fieldset
+import org.jetbrains.compose.web.dom.H1
+import org.jetbrains.compose.web.dom.Text
+import org.w3c.dom.HTMLDivElement
+
+@Composable
+fun SelectUsernameToSignInScreen(
+  attrs: AttrsScope<HTMLDivElement>.() -> Unit = {},
+  eventListener: (SignUpEvent) -> Unit,
+  options: List<UsernameSlug>,
```
15characterlimi, line 18: nit (here and elsewhere): is there a reason why you don't use Set<UsernameSlug> (and similarly for email addresses and passkeys elsewhere)?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpModel.kt
```diff
@@ -1,15 +1,37 @@
 package com.wasmo.client.app.signup
 
+import com.wasmo.identifiers.UsernameSlug
+
 data class SignUpModel(
```
15characterlimi, line 5: Should this be SignUpSignInModel? Looks like this is a mix of sign-in, sign-up actions as well as other actions that are only the first step towards signing up/in. Right?

Is it deliberate that selectUsernameToSign doesn't end in ..In or ...Up?

A one-line comment on each option below for how this method works would be really helpful. I imagine it's something like:

 - enterEmailAddressModel: enter an email address, will be emailed a challenge code to confirm, getting to that step won't yet sign in
 - newAccountWithUsername: enter a new username to sign-up and simultaneously sign-in as that username
 - selectUsernameToSign(In?): select an *existing* username, sign-in as that
 - enterChallengeCode: enter an email address and the challenge code that was recently sent to that email address

Is that about right?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SelectUsernameToSignInScreen.kt
```diff
@@ -0,0 +1,70 @@
...
+
+      if (canCreateUsername) {
+        Fieldset {
+          Button(
+            busy = busy,
+            disabled = !canSubmit,
+            outline = true,
+            attrs = {
+              onClick {
+                eventListener(SignUpEvent.ClickNewAccount)
```
15characterlimi, line 60: Should this be called ClickCreateNewUsername?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpEvent.kt
```diff
@@ -13,5 +15,16 @@ sealed interface SignUpEvent {
   data class EditChallengeCode(
     val challengeCode: String,
   ) : SignUpEvent
-}
 
+  data class ClickUsername(
```
15characterlimi, line 19: Does this correspond to SignUpModel.selectUsernameToSign? This would be more readable if names were consistent. (I'm struggling to follow)

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpModel.kt
```diff
@@ -1,15 +1,37 @@
...
+
+data class SelectUsernameToSignInModel(
+  val usernameOptions: List<UsernameSlug> = listOf(),
+  val canCreateUsername: Boolean = false,
+  val canSubmit: Boolean = false,
+)
 
+data class NewAccountWithUsernameModel(
+  val username: String = "",
+  val usernameCaption: String = "",
```
15characterlimi: What are the captions for? Is it for a status message like "available" or "already taken"? Might be more readable with a one-line comment.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpUi.kt
```diff
@@ -22,27 +22,62 @@ class SignUpUi(
   ) {
     val state by presenter.model.collectAsState()
 
-    if (state.challengeCodeEmailAddress == null) {
+    val enterChallengeCodeModel = state.enterChallengeCode
+    if (enterChallengeCodeModel != null) {
```
15characterlimi, line 26: no action required: 

I know you did this deliberately, so it's fine, but it still feels weird to me to have all these options as siblings next to each other rather than a one-off-these (sealed class or similar).

The data format makes it possible for the server to send:

 - more than one option - then the one that comes first in the (arbitrary) order will be picked
 - zero options - then we run into error("no screen?")

It feels weird to me that we're picking a data format that allows encoding invalid states, and relies on error prone and arbitrary decisions on the client to make sense of.

I imagine you have made experiences in the past that led you to this conviction/model. Perhaps some day I'll understand :-)

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
```diff
@@ -36,47 +38,52 @@ class SignUpPresenter(
...
+                challengeToken = challengeToken,
+              ),
             )
           }
         }
       }
 
       is SignUpEvent.EditChallengeCode -> {
         mutableModel.update {
+          val enterChallengeCode = it.enterChallengeCode ?: return // Race.
```
15characterlimi, line 70: Is it worth being more specific? I don't understand kind of Race condition these // Race. comments call attention to.

15characterlimi (commented):
hopefully my review comments are more helpful than onerous! - feel free to ignore.

os/api/src/commonMain/kotlin/com/wasmo/api/Username.kt
swankjesse: Ooooh, no using the `UsernameSlug` here is an improvement. I can’t use it in the UI itself ’cause it’s legal to put an invalid username in the text field before you press the button

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/api/RealWasmoApi.kt
swankjesse: Both endpoints should complete with the username linked to the client that called ’em. But `createUsername()` is intended to put a new row in the table, and `linkUsername` is intended to find an existing row in the table.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SelectUsernameToSignInScreen.kt
swankjesse: I want deliberate ordering and immutable more than anything. I’ve got a bunch of different bad options:
 - `LinkedHashSet` is mutable
 - `Set` doesn’t express deliberate ordering particularly well

So I go with `List`.

I could kind of do it with `LinkedHashSet` implementation and `Set` in the API. But then my tests still pass if the ordering is wrong. Unless I go out of my way!

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpModel.kt
swankjesse: Exactly right. I added the comments!

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SelectUsernameToSignInScreen.kt
swankjesse: It depends? The button label is 'New Account' even though the API is creating a username.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpEvent.kt
swankjesse: Yeah, agreed. The event names are intended to match the UI controls, but I’m inconsistent between naming the APIs and the screens.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpModel.kt
swankjesse: That’s exactly right. But the Pico docs use the phrase ‘helper text’ and I’ll switch to that.
https://picocss.com/docs/forms/input#validation-states

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpUi.kt
swankjesse: Ultimately I’m trying to make it so we don’t discard data if you press back. But I don’t have 'back' wired up yet.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SelectUsernameToSignInScreen.kt
swankjesse: The events here represent what the buttons and fields are doing, which roll up into the API calls

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
swankjesse: It’s indicating that we’re getting an event in a state where we’re not prepared to process. Probably impossible now so `!!` might be equivalent.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
15characterlimi: Heads up, I'm working on this now.

PR #37, opened 2026-05-12
DB schema: Support Username tombstoning / replacement.

15characterlimi:
This isn't wired up, it's unused and untested.

I edited schema history, so DB needs to be wiped and recreated to become consistent again.

I had this PR parked for when we have a solution to editing schema history without collaborators needing to be aware of the need to wipe their database, however none of my approaches was convincing so I'm sending this PR now just to get it out of the way.


os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/DbUsername.kt
```diff
@@ -10,4 +10,5 @@ data class DbUsername(
   val createdAt: Instant,
   val accountId: AccountId,
   val usernameSlug: UsernameSlug,
+  val tombstonedAt: Instant?,
```
swankjesse: nit: deletedAt ?

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
```diff
@@ -1,10 +1,15 @@
 CREATE TABLE Username (
   id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  -- Subject to clock limitations (e.g. may not be strictly increasing).
   created_at TIMESTAMPTZ NOT NULL,
-  -- 0 or 1 usernames per account. If this ever changes, need to drop 'UNIQUE' here and on the index.
-  account_id BIGINT UNIQUE NOT NULL,
-  username VARCHAR(255) UNIQUE NOT NULL
+  account_id BIGINT NOT NULL,
+  -- cannot be reused, even if tombstoned
```
swankjesse: why not?

swankjesse: I guess it’s probably easier to start with the data model maximally locked down for now, and we can relax it when we add new features

swankjesse (approved):

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
15characterlimi: Deleted the comment.

The comment expressed our intention. The data model doesn't enforce perfmanence of deletion, although we don't currently provide SQL that would allow un-deleting a username.

Un-deleting would weaken the invariants on account history, since it would break the chain of 

usernameA created at A, deleted at B,
usernameB created at B, deleted at C,
usernameC created at C

That said, chaining them via the timestamps is dubious anyways because clocks can move backwards in time and might show t he same time twice in a row, so e.g. there's no guarantee that A < B < C even if that was the sequence of events.

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/DbUsername.kt
15characterlimi: Replaced tombon -> delet throughout (preserving case of the t/d).

I had picked tombstoned because the DB entry remains in place, but I'm also okay with "deleted" since that is true at the abstraction layer just above the storage.

PR #39, opened 2026-05-14
Introduce accountSnapshot.isSignedIn and determine it server-side.

15characterlimi:
HomeUi now uses this instead of accountSnapshot.emailAddresses.isNotEmpty().

This is prep for supporting other ways of being signed-in.

This PR replaces https://github.com/wasmcomputercompany/wasmo/pull/34, which I have closed.


swankjesse (approved):

swankjesse:
Love it

PR #40, opened 2026-05-14
Run composite DB operations in context(sqlTransaction: SqlTransaction).

15characterlimi:
I hadn't known this was possible to express.

This also caught that ServiceTester was accidentally not using a transaction (not that it matters that much for tests).

### DIFF
--- a/os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
+++ b/os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
@@ -10,6 +10,7 @@ import okio.Path
 import okio.Path.Companion.toPath
 import wasmo.sql.SqlConnection
 import wasmo.sql.SqlException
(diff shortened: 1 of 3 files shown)


swankjesse (approved):

swankjesse:
I’m surprised you can do all of these operations in a transaction!

15characterlimi:
>  I’m surprised you can do all of these operations in a transaction!

Why couldn't you?

Obviously, I slightly dislike that there's all this business logic running inside the transaction, which makes the transaction longer and more likely to conflict. That's why I would expect the concept of a transaction to exist at the SqlConnection level, so that there can be multiple concurrent transactions per DB (using the transaction as the level of isolation - multiple concurrent transactions are okay as long as only one of them writes).

swankjesse:
There can be multiple concurrent transactions per DB. And you can use a connection with a transaction or without one.

PR #42, opened 2026-05-14
Username sign-in step: Introduce and partially wire-up SignInSnapshot.

15characterlimi:
I plan to merge this together with the follow-up PR https://github.com/wasmcomputercompany/wasmo/pull/45

---

Right now, SignInSnapshot is username-centric:
 - OsPage adds it iff a username exists in the DB.
 - SignUpPresenter uses its presence to switch between email vs.
   username sign-in

TODO:
 - Make SignInSnapshot the model controlling all sign-in options.
 - Introduce /sign-in route; usernames from DB should only be shown on
   /sign-in, not on /sign-up
 - Server-side endpoint to link selected username to account.

### DIFF
--- a/os/client/app/src/jsMain/kotlin/com/wasmo/client/app/home/HomeUi.kt
+++ b/os/client/app/src/jsMain/kotlin/com/wasmo/client/app/home/HomeUi.kt
@@ -9,6 +9,7 @@ import com.wasmo.api.ComputerListSnapshot
 import com.wasmo.api.routes.BuildYoursRoute
 import com.wasmo.api.routes.ComputerHomeRoute
 import com.wasmo.api.routes.RouteCodec
+import com.wasmo.api.routes.SignInRoute
 import com.wasmo.api.routes.SignOutRoute
 import com.wasmo.api.routes.SignUpRoute
 import com.wasmo.client.app.Environment
@@ -75,7 +76,7 @@ class HomeUi(
 
       HomeEvent.ClickSignIn -> {
         menuVisible = false
-        router.goTo(SignUpRoute, TransitionDirection.PUSH)
+        router.goTo(SignInRoute, TransitionDirection.PUSH)
       }
 
       HomeEvent.ClickSignUp -> {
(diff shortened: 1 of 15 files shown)


os/client/app/src/jsMain/kotlin/com/wasmo/client/app/home/HomeUi.kt
```diff
@@ -75,7 +76,7 @@ class HomeUi(
 
       HomeEvent.ClickSignIn -> {
         menuVisible = false
-        router.goTo(SignUpRoute, TransitionDirection.PUSH)
+        router.goTo(SignInRoute, TransitionDirection.PUSH)
```
swankjesse, line 79: I’ve got plans for the SPA to make sign-in and sign-up use the same URL. That URL might be the main marketing page, ie. wasmo.com or wasmo.local on Homelab. (We currently navigate, but that’s a shortcut and not a design)

os/server/calls/real/src/jvmMain/kotlin/com/wasmo/calls/RealCallDataService.kt
```diff
@@ -108,9 +129,9 @@ class RealCallDataService(
         nextChallenge = client.challenger.create(),
         passkeys = passkeys.get(),
         emailAddresses = emailAddresses,
+        username = linkedUsername.get(),
         hasInvite = firstInvite != null,
-        // TODO: isSignedIn should also reflect other ways of being signed in.
-        isSignedIn = emailAddresses.isNotEmpty(),
+        isSignedIn = emailAddresses.isNotEmpty() || linkedUsername.get() != null,
```
swankjesse, line 134: nice!

swankjesse (approved):
YAY

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/home/HomeUi.kt
15characterlimi: Thanks for the heads-up. I'll keep SignInRoute and SignUpRoute for now  if you don't mind, to get my PRs in, but this logic needs to change anyway and the split vs non-split routes won't make much difference re the difficulty.

15characterlimi:
Merged as part of https://github.com/wasmcomputercompany/wasmo/pull/45

PR #44, opened 2026-05-15
Connect to the database as a normal, non-superuser user

swankjesse:


swankjesse:
I believe we were hitting Postgresql superuser connection limits (3 concurrent connections). Non-superusers can have 100 concurrent connections by default, and so I think that’s what we want in both test and production.

15characterlimi:
Nice, thanks for the heads-up!

PR #45, opened 2026-05-15
Implement {Create,Link}UsernameRpc and wire up with client-side UI.

15characterlimi:
Implement {Create,Link}UsernameRpc and wire up with client-side UI.

Add a basic UsernameRpcsTest.

Notes:
 - Both /sign-in and /sign-up pages are handled by SignUpPresenter.
   Whether username is offered currently depends on whether
   signInSnapshot is in the page; OsPage.kt adds this to HomeRoute
   and SignInroute if there's a username in the database.
 - {Create,Link}UsernameRpc are implemented on top of a shared helper
   function. They also return a shared Decision type, which does
   not distinguish between LinkedExisting vs. LinkedNew.
 - There's no rate limiting.
 - Just like for email addresses, arbitrarily many usernames can be
   linked to the same account.
 - UsernameBindings is only added for Homelab. I haven't tested the other distributions.

TODO in future PRs:
 - Separate UIs for sign-up vs. sign-in. Make decision whether to show
   username based sign-in/sign-up based on configuration rather than DB.
 - When sign-up/-in fails, show an error message in the UI. I suggest
   we add `val errorMessage: String` to SignUpModel, outside of (a
   sibling to) the various sub-models.
 - After signing in, creating a computer doesn't work (expects payments
   to be set up, which doesn't make sense for Homelab).
 - Support for password-protected username sign-in.
 - I suggest we add `accountName: String?` to AccountSnapshot, to show
   in the UI. The server can set this to username or emailAddress,
   depending on how the user is signed-in.
 - I think we should probably allow underscores in usernames. We should either continue to disallow uppercase characters, or implement case insensitive equals/hashCode + conversion to lowercase when storing in the DB.

## Screenshot

image


os/api/src/commonMain/kotlin/com/wasmo/api/Username.kt
```diff
@@ -10,15 +10,28 @@ data class CreateUsernameRequest(
...
-data object LinkUsernameResponse
+data class LinkUsernameResponse(
+  val decision: UsernameDecision,
+  val account: AccountSnapshot?,
+)
+
+enum class UsernameDecision {
+  /** The caller's cookie was attached to the (new or existing) username specified in the request. */
+  Success,
+  /** Bad request, e.g. an invalid username was provided or a username is already linked. */
```
swankjesse: these are very different results! BadRequest should crash (software bug), whereas we gotta recover from AlreadyLinked

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameBindings.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.wasmo.framework.ActionRegistration
+import com.wasmo.identifiers.HostnamePatterns
+import com.wasmo.identifiers.OsScope
+import dev.zacsweers.metro.BindingContainer
+import dev.zacsweers.metro.ElementsIntoSet
+import dev.zacsweers.metro.Provides
+import dev.zacsweers.metro.SingleIn
+
+@BindingContainer
+abstract class UsernameBindings {
```
swankjesse, line 12: rad

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
```diff
@@ -0,0 +1,123 @@
...
+}
+
+private enum class Mode(val allowLinkExisting: Boolean, val allowLinkNew: Boolean) {
+  LinkExisting(allowLinkExisting = true, allowLinkNew = false),
+  LinkNew(allowLinkExisting = false, allowLinkNew = true),
+}
+
+private data class UsernameRpcResponse(val decision: UsernameDecision, val account: AccountSnapshot?)
+
+private suspend fun usernameRpc(
```
swankjesse: Could you do this polymorphically? Injecting dependencies by passing them in a function is weak

(You could create a new class `UsernameLinker` and make these fields of that)

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
```diff
@@ -0,0 +1,123 @@
...
+    if (mode.allowLinkNew) {
+      insertUsername(
+        createdAt = now,
+        accountId = accountId,
+        usernameSlug = username
+      )
+      return UsernameDecision.Success // linked new username
+    }
+  } catch (_: SqlException) {
+    return UsernameDecision.BadRequest
```
swankjesse: Could you catch the specific expected conflict exception and crash in all other cases? 
See `isUniqueViolation` in `Sql.kt`

os/server/usernames/real/src/jvmTest/kotlin/com/wasmo/usernames/UsernameRpcsTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import com.wasmo.testing.service.ServiceTester
+import kotlin.test.Test
+import kotlinx.coroutines.test.runTest
+
+class UsernameRpcsTest {
+  @InterceptTest
+  val tester = ServiceTester()
+
+  @Test
+  fun `create and then link username RPCs from one client`() = runTest {
```
swankjesse, line 20: love it

swankjesse (approved):

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
15characterlimi: Will do! Amusingly, this is how I had done it originally, but then I remembered you mentioning that you only like inheritance to express hierarchy but dislike it for sharing functionality, and I wasn't sure if it would meet the bar for you :-)

Will change it back to a class before I merge this PR.

os/api/src/commonMain/kotlin/com/wasmo/api/Username.kt
15characterlimi: Ack, I will reserve BadRequest for programming bugs (e.g. serialized username is empty or contains invalid characters), and will introduce a new code will have to think about the possible cases in order to figure out what code(s) to introduce.

Perhaps InvalidStateForThisUsername (for when a username is unexpectedly already present, or unexpectedly not present yet in the DB). We'll see.

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
15characterlimi: Added code above to recognize the case of the username already existing, so that I can do

```
    return if (e.isUniqueViolation || e.isDuplicateDatabase) {
      UsernameDecision.InternalServerError
    } else {
      UsernameDecision.BadRequest
    }
```
 
here.

os/api/src/commonMain/kotlin/com/wasmo/api/Username.kt
15characterlimi: Introduced:

```
  /**
   * The chosen username exists but has been deleted and cannot be reused.
   */
  UsernameDeleted,
  /**
   * The client asked to link a new username but it already exists (and wasn't
   * deleted), or to link an existing username but it doesn't exist (not even in
   * deleted form).
   */
  UsernameExistenceIncompatibleWithRequest,
```

15characterlimi: Also introduced `InternalServerError` to distinguish from `BadRequest`'s client-side errors.

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
15characterlimi: When you say "crash", do you mean rethrow the exception and let higher-level logic determine what to do?

I guess that actually makes more sense, because there must be higher-level logic that catches uncaught exceptions anyway.

I'll drop InternalServerError outcome again and instead just rethrow the exception.

15characterlimi: P.S. I *hope* that there is higher-level logic that translates uncaught exceptions

 - parsing the request (e.g. if the client sends a string that causes UsernameSlug ctor to throw IllegalArgumentException) into HTTP 4xx status codes
 - from handling the request (e.g. SqlException) into HTTP 5xx status codes
 
I don't know if this code exists and where it would live, but if it doesn't exist, we should add it.

os/api/src/commonMain/kotlin/com/wasmo/api/Username.kt
15characterlimi: Nevermind, dropped InternalServerError again and am instead letting higher-level code hopefully translate the uncaught exception into a HTTP 5xx status code.

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
15characterlimi: Done.

I split UsernameLinker into an interface + RealUsernameLinker implementation, but because it's only used by the (real) UsernameRpcs as opposed to being part of the module's API, I didn't bother introducing usernames/api but kept both in usernames/real.

15characterlimi:
Merged via rebase onto `main`, including the two commits from the base PR https://github.com/wasmcomputercompany/wasmo/pull/42 that you had already approved.

To give you the option (not expectation!) to review what changes I made in response to your review comments, I kept "address review comments" as a separate commit when rebasing: https://github.com/wasmcomputercompany/wasmo/commit/fb78bc1054e8a8f17928fbd9a542ad87219b5d28

PR #46, opened 2026-05-15
Clean up our Absurd job queue after each test

swankjesse:


os/server/jobs/absurd/src/jvmMain/kotlin/com/wasmo/jobs/absurd/AbsurdOsJobQueue.kt
```diff
@@ -41,7 +43,13 @@ internal class AbsurdOsJobQueue<P : Any, R : Any> private constructor(
...
     sqlTransaction.afterCommit {
       factory.scope.launch {
-        factory.absurdService.absurd.executeBatch("AbsurdOsJobQueue")
+        try {
+          factory.absurdService.absurd.executeBatch("AbsurdOsJobQueue")
+        } catch (_: VertxException) {
+          // Pool closed?
+        } catch (_: RejectedExecutionException) {
+          // Pool closed?
+        }
```
swankjesse, line 52: This stuff is pretty ugly. I wanna move automatic runBatch into something that only triggers in tests, so we don’t hide problems in real code

PR #47, opened 2026-05-15
Fix incorrect docker command.

15characterlimi:
See https://docs.docker.com/reference/cli/docker/container/rm/


swankjesse (commented):
Thank you!

PR #48, opened 2026-05-16
Run all JVM tests in Buildkite

swankjesse:
This creates a new Dockerfile that sets up all of our requirements:
 - wabt
 - gradle
 - node
 - JVM

It also creates a docker-compose file that sets up Postgresql.


swankjesse:
Check it out: 
https://hub.docker.com/r/wasmo/ci

Ooooooh
https://buildkite.com/wasmo/build/builds/272/tests

PR #50, opened 2026-05-16
selectLinkedUsernameOrNull(): get rid of allowDeleted: Bool parameter

15characterlimi:
- split `selectLinkedUsernameOrNull()` into two methods
- since the version that allows deleted usernames returns at most one result, implement the version that doesn't on top of it and filter out a potentially deleted username at the Kotlin rather than postgresql level.

Returning deleted usernames is normally the surprising case, therefore the function that does has a slightly unwieldy `AllowDeleted` suffix; in this particular case though, it's the only one of the two functions that's currently being used.

### DIFF
--- a/os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
+++ b/os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
@@ -72,7 +72,6 @@ suspend fun replaceUsername(
 
 /**
  * Marks [username] for the given [accountId] as deleted at [deletedAt].
- * A deleted username will no longer be found by [selectUsernameOrNull].
  * Deletion is meant to be permanent, but we might some dayy want to offer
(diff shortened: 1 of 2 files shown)


os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
```diff
@@ -123,7 +122,12 @@ suspend fun findUsernameOrNull(
 context(connection: SqlConnection)
 suspend fun selectLinkedUsernameOrNull(
   username: UsernameSlug,
-  allowDeleted: Boolean = false,
+): DbUsername? =
+  selectLinkedUsernameOrNullAllowDeleted(username)?.takeIf { it.deletedAt == null }
+
+context(connection: SqlConnection)
+suspend fun selectLinkedUsernameOrNullAllowDeleted(
```
swankjesse, line 129: good name

swankjesse (approved):

PR #51, opened 2026-05-16
Split UsernameDecision -> {Create/Link}UsernameDecision

15characterlimi:
This PR builds on top of https://github.com/wasmcomputercompany/wasmo/pull/51 , which is probably uncontroversial.

---

Flatten / inline code that was previously shared between the two cases in
RealUsernameLinker.

 - Because there's no longer any shared logic, the UsernameLinker
   abstraction was no longer pulling its weight. I inlined the code back
   into UsernameRpcs.kt. This saves a lot of boilerplate.
 - The BadRequest case was essentially unreachable. Code that tried to
   distinguish 5xx vs. 4xx cases of SqlException wasn't worth it. I've
   dropped both.

Question: Should we rename Decision -> Outcome, Verdict, Result?


swankjesse (approved):
It’s perfect

15characterlimi:
FYI in case you missed it, this PR was set to merge into the base PR's branch (rather than to main), but I think you had already merged that base PR so the extra commit on the base PR's branch never made it to `main`.

I've now manually cherry-picked this PR's commit and pushed it to `main` as https://github.com/wasmcomputercompany/wasmo/commit/f5bc32efb3917ca215438849621ff1b2f7ce4954

I hope that that's what you would have wanted me to do!

PR #53, opened 2026-05-20
 Package Homelab as a standalone Docker image 

swankjesse:


swankjesse:
This PR does too much and I don’t think it’s exemplar for contributors to follow. I should have done a separate PR for adopting Clikt for argument parsing in the three main functions.

And I don’t believe this is sufficient to actually do what we want, as we don’t yet run the commands to create the `wasmo_homelab` DB.

os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/OsDatabaseInitializer.kt
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Initializes the OS database.
+ */
+class OsDatabaseInitializer(
+  private val clientFactory: PostgresqlClient.Factory,
+  private val superuser: PostgresqlAddress,
+  private val address: PostgresqlAddress,
+) {
+  suspend fun initialize() {
```
swankjesse, line 13: The next step is to call this from `ExecPostgresqlOperator`, or near there.

PR #55, opened 2026-05-20
Allow uppercase characters and limited punctuation in usernames.

15characterlimi:
Uppercase characters and punctuation are preserved, but ignored for concerns of uniqueness constraint, `equals`, and `hashCode`. This is done by deriving a `normalized value` on top of which these concerns are expressed.

To guarantee consistency between Kotlin and PostgreSQL concepts of equality, the normalized value is stored explicitly in a DB column, rather than using a [ICU Custom Collation](https://www.postgresql.org/docs/current/collation.html#ICU-CUSTOM-COLLATIONS). This means:

 - If we ever change our normalization algorithm, we'll need a DB migration step to update the stored normalized values; for now, we check for consistency at read time and throw SqlException upon violation.
 - Developers need to wipe their database to adopt the latest schema.

### Properties of the new design

 - These characters count as punctuation: `-`, `_`, `.`.
 - The normalized value is converted to lowercase using the root Locale and punctuation is stripped, i.e. the following usernames are all equal: `johndoe`, `John.Doe`, `john-doe`, `john_doe`, `j.o.h.n_d-o-e`.
 - Punctuation may not appear in the first or last position. Thus, the normalized form will never be empty.
 - Not more than one punctuation character may appear in a row.
 - Sequences of digits are still allowed, except in the first position.

### Notes

 - No other slugs (AppSlug, ComputerSlug, DatabaseSlug) allow these characters. Should any of them?
 - I also changed length 1-15 -> 3-20 characters (before normalization, which implies 2-20 characters after normalization).
 - I added logic to preseve the case / punctuation of the value the user potentially entered at the login prompt, but right now this logic isn't used because our UI only allows the user to select their username from already-canonical list.

### Screenshot
Screenshot From 2026-05-21 01-01-14

### DIFF
--- a/identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
+++ b/identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
@@ -3,20 +3,32 @@ package com.wasmo.identifiers
 import kotlin.jvm.JvmInline
 import kotlinx.serialization.Serializable
 
-@JvmInline
+/**
+ * A valid username. All characters allowable in a username are URL safe.
+ */
 @Serializable
+@JvmInline
 value class UsernameSlug(val value: String) {
+
   init {
-    require(value.matches(UsernameSlugRegex)) {
+    require(isUsernameValid(value)) {
       "invalid username: $value"
     }
   }
 
-  companion object {
-    fun of(unnormalizedValue: String) = UsernameSlug(unnormalizedValue.lowercase())
-  }
-
+  // DB required normalizedValue to be UNIQUE among usernames
+  val normalizedValue: String
+    get() = value.replace(Regex("[._-]"), "").lowercase()
 }
 
-/** Between 1 and 15 lowercase (for case insensitivity) letters or digits, and the first is not a digit. */
-val UsernameSlugRegex = Regex("[a-z][a-z0-9]{0,14}")
+/**
+ * 3-20 characters that are alphanumeric or limited punctuation `[._-]`. No punctuation in first or
+ * last position, and not more than one in a row. No digits [0-9] in the first position.
+ */
+private val UsernameSlugRegex = Regex("^[a-zA-Z][a-zA-Z0-9._-]+[a-zA-Z0-9]$")
+
+fun isUsernameValid(usernameString: String): Boolean =
+  // Check the length first rather than get DoSed matching the regex against a very long String
+  usernameString.length in 3..20
+    && UsernameSlugRegex.matches(usernameString)
+    && Regex("[._-]{2}") !in usernameString
--- a/identifiers/src/commonTest/kotlin/com/wasmo/identifiers/UsernameSlugTest.kt
+++ b/identifiers/src/commonTest/kotlin/com/wasmo/identifiers/UsernameSlugTest.kt
@@ -1,0 +1,136 @@
+package com.wasmo.identifiers
+
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import assertk.assertions.isFalse
+import assertk.assertions.isNotEqualTo
+import assertk.assertions.isTrue
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class UsernameSlugTest {
+
+  @Test
+  fun `Usernames must be between 3 and 20 characters long`() {
+    assertThat(isUsernameValid("")).isFalse()
+    assertThat(isUsernameValid("a")).isFalse()
+    assertThat(isUsernameValid("ab")).isFalse()
+
+    assertThat(isUsernameValid("aac")).isTrue()
+    assertThat(isUsernameValid("aaaa")).isTrue()
+    assertThat(isUsernameValid("a".repeat(19))).isTrue()
+    assertThat(isUsernameValid("a".repeat(20))).isTrue()
+
+    assertThat(isUsernameValid("a".repeat(21))).isFalse()
+    assertThat(isUsernameValid("a".repeat(22))).isFalse()
+    assertThat(isUsernameValid("a".repeat(10000))).isFalse()
+  }
+
+  @Test
+  fun `Uppercase and punctuation characters are preserved but stripped from DB form`() {
+    val username = UsernameSlug("J_o_h_n.D-o-e")
+    assertEquals("J_o_h_n.D-o-e", username.value)
+    assertEquals("johndoe", username.normalizedValue)
+  }
+
+  @Test
+  fun `No punctuation characters in first or last position`() {
+    assertThat(isUsernameValid("-JohnDoe")).isFalse()
+    assertThat(isUsernameValid("_JohnDoe")).isFalse()
+    assertThat(isUsernameValid(".JohnDoe")).isFalse()
+    assertThat(isUsernameValid("JohnDoe-")).isFalse()
+    assertThat(isUsernameValid("JohnDoe_")).isFalse()
+    assertThat(isUsernameValid("JohnDoe.")).isFalse()
+
+    // For comparison, these are valid
+    assertThat(isUsernameValid("John.Doe")).isTrue()
+    assertThat(isUsernameValid("John.doE")).isTrue()
+  }
+
+  @Test
+  fun `No more than one punctuation character in a row`() {
+    assertThat(isUsernameValid("John--Doe")).isFalse()
+    assertThat(isUsernameValid("John__Doe")).isFalse()
+    assertThat(isUsernameValid("John...Doe")).isFalse()
+    assertThat(isUsernameValid("John-_Doe")).isFalse()
+    assertThat(isUsernameValid("John._Doe")).isFalse()
+    assertThat(isUsernameValid("John-._Doe")).isFalse()
+
+    // For comparison, these are valid
+    assertThat(isUsernameValid("John.Doe")).isTrue()
+    assertThat(isUsernameValid("J-o-h-n_D.o.e")).isTrue()
+    assertThat(isUsernameValid("John_Doe")).isTrue()
+  }
+
+  @Test
+  fun `Sequences of digits 0-9 are allowed except in the first position`() {
+    assertThat(isUsernameValid("jesse123")).isTrue()
+    assertThat(isUsernameValid("jesse.123")).isTrue()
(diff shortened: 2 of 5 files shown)


identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
```diff
@@ -1,22 +1,45 @@
...
     }
   }
 
-  companion object {
-    fun of(unnormalizedValue: String) = UsernameSlug(unnormalizedValue.lowercase())
-  }
+  val normalizedValue: String =
+    value.replace(Regex("[._-]"), "").lowercase()
+
+  override fun equals(other: Any?): Boolean = other is UsernameSlug && normalizedValue == other.normalizedValue
```
swankjesse: Ooooh I don’t like the consequences of this. I think objects should only be equals if they’re interchangeable, and if we want to preserve case we need them to be non-equal. (We can call `a.normalizedValue  == b.normalizedValue` explicitly when we want that behavior).

As-is, unit tests will incorrectly pass if we break case preserving behavior.

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
```diff
@@ -1,22 +1,45 @@
...
+
+  override fun hashCode(): Int = normalizedValue.hashCode()
 
+  override fun toString(): String = value
+
+  /** Returns the given preferred value, or throws if the preferred value `!= this`. */
+  fun adoptPreferredEquivalent(equalUsername: UsernameSlug) = equalUsername.also {
```
swankjesse: I don’t think this function name works for me

swankjesse: It looks like a side-effecting function, but it’s just a check

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
```diff
@@ -1,22 +1,45 @@
...
+  }
 }
 
-/** Between 1 and 15 lowercase (for case insensitivity) letters or digits, and the first is not a digit. */
-val UsernameSlugRegex = Regex("[a-z][a-z0-9]{0,14}")
+/**
+ * 3-20 characters that are alphanumeric or limited punctuation `[._-]`. No punctuation in first or
+ * last position, and not more than one in a row. No digits [0-9] in the first position.
+ *
+ * Explanation of regex:
```
swankjesse: I think this regex explanation is working very hard in a way that source code wouldn’t have to.

```
fun isUsernameValid(usernameString: String) {
  return UsernameSlugRegex.matches(usernameString) 
    && usernameString.length in 3..20
    && Regex("[._-]{2}") !in usernameString
}
```

identifiers/src/commonTest/kotlin/com/wasmo/identifiers/UsernameSlugTest.kt
```diff
@@ -0,0 +1,143 @@
...
+    assertThat(isUsernameValid("qüestionable")).isFalse()
+
+    // Other than the illegal characters, it would have been fine
+    assertThat(isUsernameValid("questionable")).isTrue()
+  }
+
+  // If the normalization algorithm changes, we need a DB migration to persisted normalized
+  // values.
+  @Test
+  fun `Normalization algorithm has not changed`() {
```
swankjesse, line 114: neat

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
```diff
@@ -145,7 +150,8 @@ suspend fun selectLinkedUsernameOrNullAllowDeleted(
     bindUsername(0, username)
   }
   return rowIterator.singleOrNull {
-    getUsername()
+    // If the given vs persisted username's non-normalized values differ, preserve the given one.
```
swankjesse: I’m not ready for our `Queries` classes to have smarts like this. I‘d like to limit this to stuff that the DB can do, so calling code is always responsible for interpreting and processing what the DB has.

(this is for dependency-injection reasons, and to force the `db` classes to stay small.)

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
```diff
@@ -174,11 +181,21 @@ suspend fun findUsernamesThatCanSignIn(
...
+    username = UsernameSlug(getString(3)!!),
+    deletedAt = getInstant(5)
+  ).also {
+    val normalizedValue = getString(4)
+    if (normalizedValue != it.username.normalizedValue) {
+      // This indicates a programming error on the Wasmo side.
+      // It means that our normalization algorithm has changed and that we ought to create a DB
+      // migration step that updates the normalized_values of the usernames in the DB
+      // (perhapsy by copying them to a new table and deleting the old one).
+      throw SqlException("DB persistent state is inconsistent: Username '${it.username.value}' " +
```
swankjesse: If this is a big in our code, it should be an `IllegalStateException` and not a `SqlException`

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
swankjesse: This needs to be `normalized_value`, not `username`

swankjesse (approved):

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
15characterlimi: Changed the class to define equals/hashCode in terms of `value` (not `normalizedValue`).

This means that the enforcement that two Usernames with equal normalizedValue can exist (e.g. in a HashMap) happens exclusively in the database - Kotlin code would allow it.

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
15characterlimi: Dropping this function.

For now, I'll just read the case-preserving username from the DB (so if a user edits the sign-in HTML to change the case of a username they log in with, they still get the original case from when the account was created).

If we later explicitly allow specifying username case during sign-in, we can then decide how to persist that choice for the session (in the cookie, update the DB, ...).

identifiers/src/commonMain/kotlin/com/wasmo/identifiers/UsernameSlug.kt
15characterlimi: Changed.

The regex was probably faster :-)

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
15characterlimi: Removed the smarts. We now always return what's in the DB, regardless of the case of what was given.

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
15characterlimi: good call, done

os/server/db/src/jvmMain/resources/migrations/v4__Username.sql
15characterlimi: dropping this index altogether, I'm pretty sure an index will implicitly be created anyway because of the UNIQUE constraint on normalized_value.

15characterlimi:
Changes in response to review comments (before then rebasing onto latest `main`): https://github.com/wasmcomputercompany/wasmo/compare/4952de586d9a6b0ba0c1011fbc8be70e3e0adb2b..4d2bb144a9f922a21f3cfa04f862e0e38c868fe4

I plan to merge this PR soon if it passes my quick manual test feel free to review the above diff and protest if you disagree with anything I changed.

15characterlimi:
The PR passed my quick manual test. However, since it would require everyone wiping their DBs again, I will hold off a little until I can either

(a) also merge https://github.com/wasmcomputercompany/wasmo/pull/59 which also requires a DB wipe, or
(b) implement some logic to auto-wipe when the .sql parrts of the migration steps has changed.

(b) is a little messy because I can't "drop database" from within a connection, I can only delete and recreate tables, indices, etc., manually. Also the .sql parts aren't everything, but it's good enough.

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
```diff
@@ -174,11 +180,21 @@ suspend fun findUsernamesThatCanSignIn(
...
+  ).also {
+    val normalizedValue = getString(4)
+    if (normalizedValue != it.username.normalizedValue) {
+      // This indicates a programming error on the Wasmo side.
+      // It means that our normalization algorithm has changed and that we ought to create a DB
+      // migration step that updates the normalized_values of the usernames in the DB
+      // (perhapsy by copying them to a new table and deleting the old one).
+      throw IllegalStateException("DB persistent state is inconsistent: Username '${it.username.value}' " +
+        "normalizes to '${it.username.normalizedValue}', but DB has '$normalizedValue'")
+    }
```
swankjesse:
```suggestion
    check(normalizedValue == it.username.normalizedValue) {
      // This indicates a programming error on the Wasmo side.
      // It means that our normalization algorithm has changed and that we ought to create a DB
      // migration step that updates the normalized_values of the usernames in the DB
      // (perhapsy by copying them to a new table and deleting the old one).
      "DB persistent state is inconsistent: Username '${it.username.value}' " +
        "normalizes to '${it.username.normalizedValue}', but DB has '$normalizedValue'"
    }
```

swankjesse (approved):

os/server/db/src/jvmMain/kotlin/com/wasmo/db/usernames/UsernameQueries.kt
15characterlimi: Thanks! My Java muscle memory is strong, I'm still acquiring idiomatic Kotlin :-)

PR #61, opened 2026-05-25
auto-wipe DB if migration steps have changed.

15characterlimi:
This is achieved by computing and comparing hash of the migration SQLs. This helps us move faster during development. This is not suitable for production, where we should have lossless migrations.

This is good enough, but may fail in extremely rare cases:

 - False positives: Custom migration steps is Kotlin are not considered, so we won't detect if those change.
 - False negatives: Whitespaces changes, hash collisions.

This change duplicates some SQL from testing code. This could potentially be de-duplicated, but that'd require a shared dependency for the SQL to live in.

Tested (on Homelab only):

 - Running this code on top of https://github.com/wasmcomputercompany/wasmo/pull/59 leads to DB wipe
 - Running it a second time does not wipe a second time.

Optional TODO for a trivial follow-up commit:

 - Should we also auto-wipe DB on schema downgrade?


os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabMigratorBindings.kt
```diff
@@ -0,0 +1,34 @@
...
+import kotlin.time.Clock
+
+@BindingContainer
+abstract class HomelabMigratorBindings {
+  companion object {
+    @Provides
+    @SingleIn(OsScope::class)
+    fun provideRealHomelabMigrator(clock: Clock, factory: RealMigrator.Factory) : Migrator =
+     factory.create(
+       customPostUpgradeSteps = mapOf(
```
swankjesse, line 22: I am anxious about `*Bindings` classes getting more behavior. These things are not test-friendly, so putting logic in 'em is dangerous. 

I don’t have a concrete alternative to suggest!

os/server/db/src/jvmMain/kotlin/com/wasmo/db/schemaversion/SchemaVersionQueries.kt
```diff
@@ -27,45 +29,51 @@ suspend fun getSchemaVersion(): DbSchemaVersion {
...
-    bindS32(1, DB_SCHEMA_VERSION_ID)
+    bindBytes(1, migrationHash)
+    bindS32(2, DB_SCHEMA_VERSION_ID)
   }
 }
 
-context(connection: SqlConnection)
-private suspend fun ensureSchemaVersionStored() {
+context(transaction: SqlTransaction)
+private suspend fun ensureSchemaVersionStored(versionZeroMigrationHash: ByteString) {
```
swankjesse, line 43: What does `versionZero` mean here?

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
```diff
@@ -40,47 +71,69 @@ open class RealMigrator(
...
+        "GRANT ALL ON SCHEMA absurd TO public"
+      )) {
+        sqlTransaction.execute(sql)
+      }
+
+      oldDbVersion = getOrCreateSchemaVersion(versionZeroMigrationHash)
+      oldVersion = oldDbVersion.version
+      if (oldVersion != 0L) {
+        throw IllegalStateException("Unexpected DB schema version after DB wipe: $oldVersion")
+      }
```
swankjesse:
```suggestion
      check(oldVersion == 0L) {
        "Unexpected DB schema version after DB wipe: $oldVersion"
      }
```

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
```diff
@@ -40,47 +71,69 @@ open class RealMigrator(
...
+    }
+
     if (oldVersion > targetVersion) {
```
swankjesse, line 125: Shouldn’t be a SqlException; that has to originate in Postgresql

swankjesse: IllegalStateException instead?

swankjesse (approved):

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabMigratorBindings.kt
15characterlimi: I think the root issue here is that half the PR (including the assisted inject stuff here) is just because I wanted to have a logger inside RealMigrator. Perhaps I overlooked a simpler way?

os/server/db/src/jvmMain/kotlin/com/wasmo/db/schemaversion/SchemaVersionQueries.kt
15characterlimi: it means version 0 of the DB revisions, corresponding to an empty series of migration steps (contains only the DatabaseSchemaVersion table, no other tables).

I considered calling this emptyMigrationHash, I think the reason I didn't go with it is because https://github.com/wasmcomputercompany/wasmo/blob/212685ab0d8e63d6af2e8f28746f4e505f56bffe/os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt#L93 and line 117 in the same file already know that versions start from 0.

Would you prefer the name emptyMigrationHash?

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
15characterlimi: Thanks, have applied this edit locally. I'll try to pick up this Kotlin idiom for the future!

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
15characterlimi: Done. Perhaps we should instead silently wipe the DB for downgrades? Then together with a `require(targetVersion >= 0)` after line 91 (just added), oldVersion > targetVersion would become impossible.

Either way I'll update https://github.com/wasmcomputercompany/wasmo/blob/c73c306236ce6ecbf68edb2e6ab0e0398d95ed09/docs/platform/plans/db_schema_migrations.md#L32

os/server/db/src/jvmMain/kotlin/com/wasmo/db/schemaversion/SchemaVersionQueries.kt
15characterlimi: P.S. The reason I'm passing this in here is that the Migrator is responsible for computing the hashes, but SchemaVersionQueries is responsible for recording schema version + hash in the database during `ensureSchemaVersionStored()` (if the version is not yet recorded so we now record version 0).

I looked into whether ensureSchemaVersionStored() should be split into nullable read vs. write steps, but that would lose us the separation where ensureSchemaVersion `INSERT`s the value whereas other commands only `UPDATE` it.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabMigratorBindings.kt
swankjesse: I think we can do a multibinding. Perhaps in a pairing session?

os/server/db/src/jvmMain/kotlin/com/wasmo/db/schemaversion/SchemaVersionQueries.kt
swankjesse: Thanks for the explanation!

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
swankjesse: We’ll soon want to halt instead of wipe when those are our only options. We’re going to have real users someday and our defaults need to be safe, even if they make it more annoying to hack on this thing.

swankjesse (approved):

os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt
15characterlimi: Sounds good, I'll not wipe on downgrade then. Downgrade should be pretty rare even during development, basically only in the rare case where we add a new version to begin with coupled with changing to a git branch that doesn't have it yet. Manual wipe is ok in that case.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabMigratorBindings.kt
15characterlimi: I'm always happy to pair etc! And I'm still learning Metro, so that's extra good.

PR #62, opened 2026-05-25
 Use specific ports in the container homelab

swankjesse:
54400 is the HTTP port.


os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
```diff
@@ -30,7 +31,7 @@ import wasmo.sql.SqlDatabase
 
 class HomelabCommand : CliktCommand() {
   val container: Boolean by option("--container")
-    .flag("--no-container", default = false)
+    .flag("--no-container", default = true)
```
swankjesse, line 34: Heads up ! We’re gonna need to start passing `--no-container` when running the Homelab outside of its Dockerfile.

PR #63, opened 2026-05-25
Fix compilation of `./gradlew os:distributions:homelab:run`

15characterlimi:
Give RealOsDatabaseInitializer's clientFactory argument a default value, since there's really only one reasonable value (?).

As of commit 5afbf8a71415da2860ad52a9525fcb23ebb54897, the command failed. Perhaps this exact command is now obsolete, because of Docker?

### DIFF
--- a/os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/PostgresqlClient.kt
+++ b/os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/PostgresqlClient.kt
@@ -1,5 +1,8 @@
 package com.wasmo.sql
 
+import com.wasmo.identifiers.OsScope
+import dev.zacsweers.metro.Inject
+import dev.zacsweers.metro.SingleIn
 import io.vertx.pgclient.PgBuilder
 import io.vertx.pgclient.PgConnectOptions
 import io.vertx.pgclient.SslMode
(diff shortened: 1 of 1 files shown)


os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/RealOsDatabaseInitializer.kt
```diff
@@ -8,8 +8,8 @@ import io.vertx.pgclient.PgException
 @Inject
 @SingleIn(OsScope::class)
 class RealOsDatabaseInitializer(
-  private val clientFactory: PostgresqlClient.Factory,
   private val config: WasmoPostgresqlConfig,
+  private val clientFactory: PostgresqlClient.Factory = PostgresqlClient.Factory(),
```
swankjesse: I don’t like default values for injector-provided things?

swankjesse (approved):

os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/RealOsDatabaseInitializer.kt
swankjesse: (Please drop the default)

15characterlimi: Okay, will mark the PostgresqlClient.Factor ctor as `@Inject` instead.

15characterlimi: Ah, you already did this on main. Sorry about the delay. I'll drop this PR instead.

PR #66, opened 2026-05-25
Return the base type, as that's what the caller expects

swankjesse:


os/server/computers/real/src/jvmMain/kotlin/com/wasmo/computers/free/FreeCreateComputerSpecRpc.kt
```diff
@@ -25,18 +24,17 @@ class FreeCreateComputerSpecRpc(
   private val wasmoDb: SqlDatabase,
   private val computerSpecStore: ComputerSpecStore,
   private val computerStore: ComputerStore,
-  private val clock: Clock,
-) : RpcAction<CreateComputerSpecRequest, CreateComputerSpecResponse.Success> {
+) : RpcAction<CreateComputerSpecRequest, CreateComputerSpecResponse> {
   suspend fun create(
     request: CreateComputerSpecRequest,
-  ): Response<CreateComputerSpecResponse.Success> {
```
swankjesse, line 32: So when this used to return `CreateComputerSpecResponse.Success`, the serialized JSON was `{}` because no type discriminator is necessary. By changing it to return the supertype, the serialized JSON gets type information injected: `{"type":"Success"}`.

15characterlimi (approved):
Thanks! I did not realize that this would affect how the value was going to be serialized to JSON.

15characterlimi (approved):
test changes LGTM - that's the downside of making each implementation return the base type

PR #68, opened 2026-05-28
Experiment with Chicory as our Wasm runtime

swankjesse:
One awesome thing about Chicory is we can skip installing wabt in our containers,because Chicory distributes them as Wasm tools: https://chicory.dev/docs/usage/tools


os/server/testing/src/jvmMain/kotlin/com/wasmo/testing/WatCompiler.kt
swankjesse, line 1: Rather than shelling out to wat2wasm, the awesome Chicory people have a Java library that calls it for us. And that works by running wat2wasm itself in Chicory!

wasmo-build/ci/Dockerfile
```diff
@@ -21,11 +21,6 @@ RUN set -o errexit -o nounset \
       tar \
       xz \
       \
-    && echo "Installing unstable utilities with apk" \
-    && apk add --no-cache \
-       --repository=https://dl-cdn.alpinelinux.org/alpine/edge/testing \
-      wabt \
```
swankjesse, line 27: wabt is no longer required!

PR #72, opened 2026-06-01
Support println() from guest code

swankjesse:


os/server/wasm/real/src/jvmTest/kotlin/com/wasmo/wasm/HostBridge.kt
```diff
@@ -67,6 +109,32 @@ class HostBridge {
...
+    val buffer = Buffer()
+
+    var iovsAddress = iovs
+    for (i in 0 until iovsSize) {
+      val sliceAddress = memory.readInt(iovsAddress)
+      iovsAddress += 4
+      val sliceSize = memory.readInt(iovsAddress)
+      iovsAddress += 4
+
+      buffer.write(memory.readBytes(sliceAddress, sliceSize))
```
swankjesse, line 130: This allocates a `ByteArray` that I’d prefer it not. I think we need to add a method to Chicory’s `Memory` to read into a caller-provided array.

PR #73, opened 2026-06-01
Start to model Wit in Wasmo

swankjesse:


jingibus:
👀👀👀👀👀

support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/Declarations.kt
```diff
@@ -0,0 +1,203 @@
...
+  val unstable: Identifier? = null,
+)
+
+data class Location(
+  val line: Int,
+  val column: Int,
+)
+
+/**
+ * Declarations may be [Use], [Interface], [World].
```
jingibus, line 26: Docs are off here

jingibus: Sorry for being overly brief — in more detail, I meant that [Use], [Interface], and [World] didn't appear to be inclusive of everything a [Declaration] might be.

But having sat down and read more closely, I'm wrong — it's accurate, it's just that the `Declaration` type doesn't map to which declarations are valid at the toplevel of a wit file

15characterlimi (commented):
This code looks like it probably is the result of fairly mechanical transcription of a WIT grammar or other piece of documentation.

Do you mind specifying what piece of documentation you followed in creating this commit?

I would have guessed https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md but I don't see a clear 1:1 mapping to this commit.

Providing this context may be helpful to readers and git archaeologists.

15characterlimi:
> This code looks like it probably is the result of fairly mechanical transcription of a WIT grammar or other piece of documentation.
> 
> Do you mind specifying what piece of documentation you followed in creating this commit?
> 
> I would have guessed https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md but I don't see a clear 1:1 mapping to this commit.
> 
> Providing this context may be helpful to readers and git archaeologists.

Nevermind, you hinted at it in support/wit/README.md; I would have expected this context to be given in the commit message and PR description (consider copying the commit message to the PR description when there's a single commit).

support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/Declarations.kt
swankjesse: Yeah, I considered trying to get the type system to enforce this. But I think I’d rather enforce it with code, because I think that’ll allow me to do better error reporting. I’ll see if this comes true.

swankjesse:
Yeah, the README gives you a sense of what I’m attempting here, and you’ll be disoriented going through this without seeing that first!

PR #74, opened 2026-06-01
Start to parse wit structure

swankjesse:


support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitStructureReader.kt
```diff
@@ -0,0 +1,120 @@
+@file:OptIn(ExperimentalContracts::class)
+
+package com.wasmo.support.wit
+
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+
+internal class WitStructureReader(
+  private val chars: CharArray,
```
15characterlimi, line 9: Consider appending a \n if chars doesn't already end in \n.

That way, you don't have to treat the case of the file not ending in a \n specially, i.e. logic like that in line 60 doesn't break.

Or do we get this for free anyway (eg. okio guarantees it)? Does okio also guarantee newline normalization?

support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitStructureReader.kt
```diff
@@ -0,0 +1,120 @@
...
+    contract {
+      returns() implies value
+    }
+    if (!value) {
+      throw WitException(location, message())
+    }
+  }
+}
+
+internal fun CharArray.indexOf(fromIndex: Int, substring: String): Int {
```
15characterlimi, line 110: nit: Consider swapping the parameter order to be consistent with StringBuilder.indexOf(String, Int)

(if you'd used StringBuilder instead of CharArray then you wouldn't have needed to implement this)

15characterlimi (commented):
nice code! Some small nits below

support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitStructureReader.kt
swankjesse: I think I wanna keep it as-is, if only because appending an extra newline seems... dishonest?

support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitStructureReader.kt
swankjesse: YES! Done in the next PR.

support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitStructureReader.kt
15characterlimi: But then the indexOf(..., "\n") in line 60 might return -1 and that code will break?

swankjesse: Yikes. Yep, nice catch.
https://github.com/wasmcomputercompany/wasmo/pull/76

PR #79, opened 2026-06-02
Parse types in Wit

swankjesse:


support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitStructureReader.kt
```diff
@@ -75,6 +83,15 @@ internal class WitStructureReader(
     return true
   }
 
+  fun tryReadLiteral(literal: String): Boolean {
```
15characterlimi, line 86: nit: Not important, but again if chars were a String or StringBuilder then this is another method you wouldn't have needed to implement, the check part could be:

sb.subSequence(pos,min(sb.length(), pos + literal.length)).contentEquals(literal)

PR #82, opened 2026-06-02
Add missing param for local ci execution

jingibus:
I suppose the other option is to just get rid of this completely


swankjesse (approved):

swankjesse:
Very handy

PR #84, opened 2026-06-03
Update local postgres setup instructions

jingibus:


docs/local_development/postgresql.md
```diff
@@ -22,6 +22,13 @@ Set up the PostgreSQL CLI:
 
 ```bash
 $ brew install libpq
+$ brew install postgresql
+```
+
+It may be necessary to create the `postgres` superuser manually:
+
+```bash
+$ createuser -h localhost -s postgres
```
swankjesse, line 32: Ooooh I assume this is necessary if you’re running Postgresql directly on your Mac, instead of via Docker.

We should probably just present those two options as alternatives, and offer two sections on setting up Postgresql. In particular, it’s definitely not necessary to do `brew install postgresql` if you’re using Docker.

swankjesse: (I’m running my Postgresql through Docker because I’m familiar with that, but it’s certainly more heavyweight the way I’m doing it.)

jingibus:
Updated per our discussion in Slack on the RIGHT way to do this - haven't confirmed it on my machine yet, but it LGTM

docs/local_development/postgresql.md
jingibus: Yeah, I did all the Docker setup to get the CI setup running locally. Is that necessary for everyone? Shrug, I dunno.

For now it's almost certainly fine, but being able to replicate CI stuff without running docker would be nice

swankjesse (approved):

swankjesse:
# 🙏🏻  thanks!

PR #85, opened 2026-06-03
Begin implementing worlds in wit

swankjesse:


support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitStructureReader.kt
```diff
@@ -21,6 +21,43 @@ internal class WitStructureReader(
   val exhausted: Boolean
     get() = pos == chars.size
 
+  /**
```
swankjesse, line 24: This function is a lot of fun, and possibly quite dangerous!

support/wit/src/jvmMain/kotlin/com/wasmo/support/wit/WitReader.kt
```diff
@@ -626,13 +645,220 @@ class WitReader private constructor(
...
+    )
+  }
+
+  /**
+   * ```ebnf
+   * export-item ::= 'export' id ':' extern-type
+   *               | 'export' use-path ';'
+   * ```
+   */
+  private fun readExport(
```
swankjesse, line 776: Imports and exports are very similar, and there’s ~25 lines of code that could be DRY'd up if we really wanted to. I’m reluctant to because I believe the implementation-similarity is not a promise of future behavior similarity, and I’d like to honor the grammar that keeps these things separate.

PR #87, opened 2026-06-03
Finish implementing our WIT reader

swankjesse:


swankjesse:
This isn’t done-done, but there’s no syntax I’m aware that I’ve missed.

I intend to do a few follow-ups:
 - decode all the `.wit` files I can find
 - review test coverage

PR #96, opened 2026-06-04
Adding a secrets wrapper and migrate to an AppDatabasePasswordSource.

mpawliszyn:


swankjesse:
Tests are failing?
```
io.vertx.pgclient.PgException: FATAL: password authentication failed for user "wasmo_test" (28P01)
io.vertx.pgclient.PgException
io.vertx.pgclient.PgException: FATAL: password authentication failed for user "wasmo_test" (28P01)
```

docs/local_development/postgresql.md
```diff
@@ -33,7 +33,7 @@ $ createuser -h localhost -s postgres
 ```
 
 And create your local databases:
-
+wasmo_homelab
```
swankjesse: I don’t think this stuff is necessary now.

https://github.com/wasmcomputercompany/wasmo/blob/main/os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/RealOsDatabaseInitializer.kt#L17

(But it’s quite likely these automatic setup would have failed for you, because your database already exists)

os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/AppDatabasePasswordSource.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.wasmo.identifiers.DatabaseSlug
+import com.wasmo.identifiers.OsScope
+import com.wasmo.identifiers.Secret
+import dev.zacsweers.metro.Inject
+import dev.zacsweers.metro.SingleIn
+
+/**
+ * Currently we just run a scheme based on os password information and the app information.
+ *
+ * We deem this safe because if an attacker happens to obtain the os password then they would
```
swankjesse, line 12: I still think we need to do the HMAC thing, because this also runs on the Hosted deployment and that thing is Internet-accessible

swankjesse (changes requested):

swankjesse:
@15characterlimi this PR introduces a schema change! What’s the best way to trigger schema updates across our other developers?

15characterlimi:
> @15characterlimi this PR introduces a schema change! What’s the best way to trigger schema updates across our other developers?

Should happen automatically, no manual action needed.

More detail:

Since you edited the existing migration step `os/server/db/src/jvmMain/resources/migrations/v1__InstalledAppDatabase.sql`, the sha256 of the migration steps will change, and therefore the first time someone runs the code after pulling your commits, the code from https://github.com/wasmcomputercompany/wasmo/pull/61 should trigger a wipe and re-creation of all DB tables etc. via this code here: https://github.com/wasmcomputercompany/wasmo/blob/5068b2a065805ce5bb94a613f36de9938a8edd56/os/server/db/src/jvmMain/kotlin/com/wasmo/db/Migrator.kt#L100-L118

---

Sorry for having been silent for the last few days BTW, I was travelling for work and have been busy, will hopefully re-engage a bit soon.

docs/local_development/postgresql.md
mpawliszyn: I wiped it and it was still broken. Adding those databases appeared to work though.

os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/AppDatabasePasswordSource.kt
mpawliszyn: Yes I was going to do that in a followup.

swankjesse (approved):

PR #99, opened 2026-06-05
Collect type references in a WIT file

swankjesse:


support/wit/wit-core/src/jvmMain/kotlin/com/wasmo/support/wit/TypeReference.kt
```diff
@@ -0,0 +1,115 @@
...
+)
+
+/** Does a depth-first traversal of [subject], looking for [TypeName] instances. */
+private class TypeReferenceScanner(
+  private val packageName: PackageName?,
+  private val interfaceName: Identifier? = null,
+  private val subject: Declaration,
+) {
+  context(sequence: SequenceScope<TypeReference>)
+  suspend fun scan() {
```
jingibus, line 40: Unfamiliar pattern to me to use a class instance rather than function params to pass this state around.

jingibus (approved):

support/wit/wit-core/src/jvmMain/kotlin/com/wasmo/support/wit/TypeReference.kt
swankjesse: Yeah, I started with params but it's really recursive and it got very busy repeating the same parameters over and over!

PR #107, opened 2026-06-07
Create a wit-testing module

swankjesse:
It's otherwise too much boilerplate to create test objects for testing higher-level features like WorldFlattener


support/wit/wit-core/src/jvmTest/kotlin/com/wasmo/support/wit/SymbolIndexTest.kt
```diff
@@ -163,36 +163,12 @@ class SymbolIndexTest {
...
-    assertThat(
-      index.getWorldOrNull(
-        UsePath(
-          namespaces = listOf(Identifier("wasi")),
-          packageNames = listOf(Identifier("cli")),
-          name = Identifier("command"),
-        ),
-      ),
-    ).isNull()
+    assertThat(index.getWorldOrNull("wasi:cli/command".toUsePath())).isNull()
```
swankjesse, line 172: This is the motivation of the PR

PR #117, opened 2026-06-11
Introduce an intermediate representation in WIT

swankjesse:
This is a big hairy refactor that changes our WIT code from having two representations (WIT and Kotlin) into having three representations (WIT IO, WIT IR, and Kotlin).

The intermediate model is unfortunate because it's duplicative. But it's also excellent because it radically simplifies all of the Kotlin-generation code, which now doesn't need to worry about resolving relative types or flattening worlds. By the time Kotlin gets to it, it's a consistent linked graph.


swankjesse:
There’s a few follow-up refactors that are necessary after this, because this code is uglier than it needs to be now that we have a new representation for the WIT data.

I haven’t done that yet, but this change is already about 1500 lines too big.

PR #118, opened 2026-06-11
Hook up OsConfig and drive kcoroutines debug with it

jingibus:
I had hoped the coroutines stuff would do more, but I think the stack trace rewriting may only come into play when the app dies.

As far as the injected value goes, I'd like to flag some other stuff with this, like verbose SQL logging, etc


swankjesse (approved):

swankjesse:
Neat!

PR #122, opened 2026-06-12
Add verbose SQL query debugging

jingibus:
Ladies and gentlemen of the PR review board... I'm just a caveman. I don't understand your modern tools.

(That is, `println` is doing the trick for me?)

### DIFF
--- a/os/api/src/commonMain/kotlin/com/wasmo/api/SqlEventListener.kt
+++ b/os/api/src/commonMain/kotlin/com/wasmo/api/SqlEventListener.kt
@@ -1,0 +1,15 @@
+package com.wasmo.api
+
+/**
+ * Listener for events triggered in the SQL service.
+ */
+fun interface SqlEventListener {
+  fun onEvent(sqlEvent: SqlEvent)
+
+  sealed interface SqlEvent {
+    /**
+     * Fired right before a SQL query or statement is executed.
+     */
+    data class SqlStarted(val query: String, val params: List<Any?>) : SqlEvent
+  }
+}
--- a/os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
+++ b/os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
@@ -6,7 +6,7 @@ import com.github.ajalt.clikt.parameters.options.flag
 import com.github.ajalt.clikt.parameters.options.option
 import com.wasmo.accounts.CookieSecret
 import com.wasmo.accounts.SessionCookieSpec
-import com.wasmo.api.OsConfig
+import com.wasmo.api.SqlEventListener
 import com.wasmo.api.stripe.StripePublishableKey
 import com.wasmo.common.catalog.DevelopmentCatalog
 import com.wasmo.identifiers.Deployment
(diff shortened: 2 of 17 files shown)


os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/RealSqlDatabase.kt
```diff
@@ -82,10 +85,20 @@ internal class RealSqlConnection(
           val preparedQuery = sqlClient.preparedQuery(sql)
           val tupleBuilder = TupleBuilder()
           tupleBuilder.bindParameters()
-          preparedQuery.execute(tupleBuilder.build())
+          val tuple = tupleBuilder.build()
+          if (debugSqlQueries) {
+            println("executing: $sql ${tuple.deepToString()}")
```
swankjesse: The pattern I’d like to establish here is events, so we can decouple ‘a thing happened’ from the effect when it happens

Here I think it’d be a new class, `SqlQueryStartEvent` 

Thoughts?

swankjesse (approved):

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
```diff
@@ -87,7 +87,28 @@ class HomelabCommand : CliktCommand() {
       )
     }
 
+    val osConfig = if (devMode) {
```
mpawliszyn: Is this enough safety for statements that create users for instance since those contain passwords?

swankjesse: Really good question. My preference is that if you’re running in the SDK or in dev mode, all privacy is gone. But the giant drawback of that is you cannot use devmode to debug issues in a live deployment.

jingibus: I might prefer to say that we just haven't built any sensitive data yet. We have passwords, but...those are just strings right now. 

If we had a distinct datatype for passwords and the like, it would be better that it never ever get logged in cleartext. That not only provides an extra layer of safety, but also makes it clear when debugging that you have a problem if you ever see a plaintext password.

os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/RealSqlDatabase.kt
jingibus: Perfect!

jingibus: Went with `SqlStarted` b/c it gets hit on both execution and query right now

os/api/src/commonMain/kotlin/com/wasmo/api/SqlEventListener.kt
```diff
@@ -0,0 +1,15 @@
+package com.wasmo.api
+
+/**
+ * Listener for events triggered in the SQL service.
+ */
+fun interface SqlEventListener {
```
swankjesse, line 6: I’d prefer to smash this event into the giant events firehose. It’s less typesafe, but it’s also got easier discoverability?

swankjesse (approved):

PR #124, opened 2026-06-12
Use an inline function in preserveStackTrace

swankjesse:


swankjesse:
This is just riffing on Bridget’s gadget!

jingibus (approved):

PR #128, opened 2026-06-12
Make the build more parallel

swankjesse:
This requires using a different DB prefix for every Gradle module.


swankjesse:
This didn’t have as much of an effect in CI as I was expecting!

PR #130, opened 2026-06-12
Update Brevity to WASIp3

swankjesse:
This uncovered a collision in some names, which needed a fix in the IR mapper.


support/brevity/brevity-kotlin-generator/src/jvmTest/kotlin/dev/wasmo/brevity/kotlin/generator/ApiGeneratorTest.kt
```diff
@@ -304,4 +304,78 @@ class ApiGeneratorTest {
...
+      """
+      |// Generated by Brevity. Do not edit.
+      |package wit.wasi.sockets.v0_3_0
+      |
+      |import dev.wasmo.brevity.Quad
+      |import kotlin.UByte
+      |import kotlin.collections.List
+      |import kotlin.jvm.JvmInline
+      |
+      |public object Imports
```
swankjesse, line 354: Note this doesn't declare a `Host` interface with the imports

PR #139, opened 2026-06-17
Move SqlEvent into global Event

jingibus:
More annoying than expected. We fire up `PostgresqlClient` before anything else, so `Logger` and `EventListener` must be made part of bootstrapping.


swankjesse (approved):
I like it

PR #143, opened 2026-06-28
Drop Homelab dependency on Stripe.

15characterlimi:
Drop Homelab dependency on Stripe via STRIPE_PUBLISHABLE_KEY and STRIPE_SECRET_KEY environment variables.

The paid vs. free distinction is now done through polymorphism in more places:

   - ComputerPaymentMethod.{Free,Stripe} replaces StripePublishableKey
     at places that at compile time might encounter either, e.g.
     RealServerOsHtml or CheckoutSession.Factory.
     - The latter produces an
       immediately-failed checkout session if invoked.
     - ComputerPaymentMethod lives in `com.wasmi.api.payments`, which
       was renamed from `*.stripe` and is consistent with
       `com.wasmo.payments`.
   - PaymentsService gains a new implementation NoPaymentsService that
     fails at runtime if used.

Notes:

 - Distributions will now depend on StripeBindings iff they use
   PaidComputersBindings (as opposed to FreeComputersBindings). We may
   want to combine the two, or more likely move the bindings of
   ComputerPaymentsMethod and PaymentsService from FreeComputersBindings
   to some free sibling of StripeBindings.
 - I experimented with separating the paid vs. free computers code more
   strictly into different compilation units, but that ballooned into
   too much boilerplate; it reduces flexibility for us to configure
   both paid and free options in the same distribution later, too.
   Not worth it.

**Tested**: Homelab only, i.e.:

`./gradlew os:distributions:homelab:run --args="--no-container"`

### DIFF
--- a/os/api/src/commonMain/kotlin/com/wasmo/api/payments/ComputerPaymentMethod.kt
+++ b/os/api/src/commonMain/kotlin/com/wasmo/api/payments/ComputerPaymentMethod.kt
@@ -1,0 +1,23 @@
+package com.wasmo.api.payments
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * How computers can be paid for.
+ */
+@Serializable
+sealed class ComputerPaymentMethod {
+  /** No payment possible - only works for computers that are free. */
+  @Serializable
+  @SerialName("payment-config-free")
+  data object FreeOnly : ComputerPaymentMethod()
+
+  /**
+   * Payment possible via Stripe - configured via StripePublishableKey, a dependency injection
+   * must be set up separately for distributions that need it.
+   */
+  @Serializable
+  @SerialName("payment-config-stripe")
+  data class Stripe(val stripePublishableKey: StripePublishableKey) : ComputerPaymentMethod()
+}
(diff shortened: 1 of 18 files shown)


os/api/src/commonMain/kotlin/com/wasmo/api/payments/ComputerPaymentMethod.kt
```diff
@@ -0,0 +1,23 @@
+package com.wasmo.api.payments
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+/**
+ * How computers can be paid for.
+ */
+@Serializable
+sealed class ComputerPaymentMethod {
```
swankjesse, line 10: Nice

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/stripe/CheckoutSession.kt
```diff
@@ -40,19 +40,25 @@ class CheckoutSession private constructor(
...
-      val stripe = Stripe(stripePublishableKey.publishableKey)
-      val deferredCheckout = stripe.initEmbeddedCheckout(
-        InitEmbeddedCheckoutOptions(
-          fetchClientSecret = { Promise.resolve(clientSecret) },
-        ),
-      ).asDeferred()
-
+      val deferredCheckout = when (computerPaymentMethod) {
+        is ComputerPaymentMethod.FreeOnly -> {
+          Promise.reject(UnsupportedOperationException("checkout not supported")).asDeferred()
```
swankjesse: A little more coroutinesy to do this
```suggestion
          CompletableDeferred<Checkout>()
          .apply { completeExceptionally(UnsupportedOperationException("checkout not supported")) }
```

(same behavior either way)

swankjesse (approved):
Nice

jingibus (approved):
🎉🎉🎉

swankjesse (approved):

PR #145, opened 2026-07-04
Spell out plan for account password storage. 

15characterlimi:
This is to ensure alignment before I start implementing. Opinions welcome.

Note: I am NOT a security expert, so this design may be boneheaded.

### DIFF
--- a/docs/code/accounts_and_authentication.md
+++ b/docs/code/accounts_and_authentication.md
@@ -44,7 +44,34 @@ Passwords are optional.
 
 Once supported, passwords can be added to accounts that do not yet have one.
 
-Passwords will be stored following [OWASP guidelines].
+Passwords will be stored following [OWASP guidelines]. Specifically we store a hash + salt for each password in the DB, see [Password Hashing](#password-hashing) below.
+
+On sign-in, passwords are only used in connection with usernames, so for now they implicitly only exist on Homelab.
+We don't support passwords for email addresses, because if the Distribution supports verify email addresses then that is enough, and if it can't then an unverified email address is just a username.
+Deliberate account sharing with a spouse works by either sharing a challenge code or password.
+
+The client is told *whether* there is a password associated with each particular username, so it knows whether to prompt for one during sign-in.
+
+At most one password per account, and only if there's a username associated with the account.
+Implementation: At most one canonical username exists per account (aliases point to it), only the canonical username may have a password.
+The pointer is transactionally updated for all aliases when a new canonical username is set for a given account.
+
+#### Password Hashing
+- We'll store an *Argon2* hash ([Wikipedia](https://en.wikipedia.org/wiki/Argon2)) for each password.
+- Inputs to the hash are:
+  - the plaintext password as UTF-8 bytes
+  - a random per-password salt (16 bytes)
+  - a random per-Wasmo instance **pepper** (TBD bytes - maybe >= 16 bytes to match salt?), passed as **associatedData** to the `Argon2` function.
+  - these hard coded values (from OWASP guidelines); the hard-coded values at the time a password is created are stored with the hashes, in case we want to change these values later.
+    - memory size m=12288 (12 MiB)
+    - iterations t=3
+    - parallelism p=1
+- Hashing happens on the server only; during login, the client sends the plaintext password over https (else the hash would become the password); open question: Is this standard, or is there a better way?
+- Protection against brute force cracking is initially only through the slowness of Argon2 + difficulty of obtaining DB + pepper. In the future, rate limiting will be added for all server calls (doesn't help protect leaked DB + pepper).
+- We'll use the [reference implementation of Argon2](https://github.com/P-H-C/phc-winner-argon2) (in C) with these [JVM bindings](https://github.com/phxql/argon2-jvm) - that project is much more active than [the alternative](https://github.com/kosprov/jargon2-api) also linked from the reference implementation GitHub page.
+- The per-Wasmo instance *pepper* value means that password hashes can only be used on that instance.
+- The pepper value is secret configuration of the Wasmo instance, like the postgresql password. It's stored in the same place: For now in an environment variable, later in a dedicated secrets vault.
+- Out of the three variants (Argon2i, Argon2d, and Argon2id), we'll ues Argon2id for hybrid protection against both classes of attack noted in the [README](https://github.com/P-H-C/phc-winner-argon2/blob/master/README.md).
 
 ### `InstanceAdminGrant`
 
--- a/docs/code/accounts_and_authentication.md
+++ b/docs/code/accounts_and_authentication.md
@@ -46,30 +46,42 @@ Once supported, passwords can be added to accounts that do not yet have one.
 
 Passwords will be stored following [OWASP guidelines]. Specifically we store a hash + salt for each password in the DB, see [Password Hashing](#password-hashing) below.
 
-On sign-in, passwords are only used in connection with usernames, so for now they implicitly only exist on Homelab.
-We don't support passwords for email addresses, because if the Distribution supports verify email addresses then that is enough, and if it can't then an unverified email address is just a username.
-Deliberate account sharing with a spouse works by either sharing a challenge code or password.
+It's the user's choice whether they want to enable password authentication for their account.
 
-The client is told *whether* there is a password associated with each particular username, so it knows whether to prompt for one during sign-in.
+- If present, the password is associated with the account (not a username), to allow flexibility to introduce passwords for accounts identified via email address. However, initially our UI might only allow setting one on Homelab (implies username).
+- It's the user's choice whether they want to choose password authentication or a more secure method.
+- At most one password per account. We don't make any efforts to discourage password sharing (e.g. hand out access keys). If you want to share your account with your spouse, you have to share the password.
 
-At most one password per account, and only if there's a username associated with the account.
-Implementation: At most one canonical username exists per account (aliases point to it), only the canonical username may have a password.
-The pointer is transactionally updated for all aliases when a new canonical username is set for a given account.
+| account identifier | has password? | how a user signs in                                         |
+|--------------------|---------------|-------------------------------------------------------------|
+| username           | password      | click on username and type password                         |
+| username           | no password   | click on username                                           |
+| email address      | password      | enter email address and password (optionally passkey)       |
+| email address      | no password   | enter email address and challenge code (optionally passkey) |
+
+ - If an account is identified by email address-identified, the client doesn't know whether the thing it's prompting the user for is a password or a challenge code.
+   - This prevents rogue clients probing whether a particular email address has an account.
+ - If an account is identifed by username, the client is told whether there's a password on the account, so it knows whether to prompt for one.
 
 #### Password Hashing
 - We'll store an *Argon2* hash ([Wikipedia](https://en.wikipedia.org/wiki/Argon2)) for each password.
 - Inputs to the hash are:
   - the plaintext password as UTF-8 bytes
   - a random per-password salt (16 bytes)
-  - a random per-Wasmo instance **pepper** (TBD bytes - maybe >= 16 bytes to match salt?), passed as **associatedData** to the `Argon2` function.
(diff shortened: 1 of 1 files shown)


docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
 
 Once supported, passwords can be added to accounts that do not yet have one.
 
-Passwords will be stored following [OWASP guidelines].
+Passwords will be stored following [OWASP guidelines]. Specifically we store a hash + salt for each password in the DB, see [Password Hashing](#password-hashing) below.
+
+On sign-in, passwords are only used in connection with usernames, so for now they implicitly only exist on Homelab.
+We don't support passwords for email addresses, because if the Distribution supports verify email addresses then that is enough, and if it can't then an unverified email address is just a username.
+Deliberate account sharing with a spouse works by either sharing a challenge code or password.
```
swankjesse: I hate passwords because they’re full of antipatterns! And I’d still like to support passwords in the hosted distribution because our users expect us to.

My personal use case for liking passwords is low-effort account sharing. My wife and I share accounts on a handful of services (Loyalty programs, Netflix, etc.). Some send an email challenge for every sign-in, and the ergonomics of that are pretty terrible because we don’t read each other’s emails. So I would like to share a password with my spouse, and be able to sign-in without asking her to forward me a challenge code.

I think our other users are better off with passkeys, but even there I think we should support passwords for auth. I’d like to be able to host a Bitwarden-like app on Wasmo, and in that case you’ll need a master password for Wasmo otherwise you could get locked out.

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+
+The client is told *whether* there is a password associated with each particular username, so it knows whether to prompt for one during sign-in.
```
swankjesse: YES! And I definitely want passwords to be optional everywhere, because it is the users’ decision if they don’t want an email challenge every time.

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+
+At most one password per account, and only if there's a username associated with the account.
```
swankjesse: Yes to one password per account
No to only-username

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+Implementation: At most one canonical username exists per account (aliases point to it), only the canonical username may have a password.
```
swankjesse: YES!

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+The pointer is transactionally updated for all aliases when a new canonical username is set for a given account.
+
+#### Password Hashing
+- We'll store an *Argon2* hash ([Wikipedia](https://en.wikipedia.org/wiki/Argon2)) for each password.
```
swankjesse, line 67: This Argon2 thing is rad. I’ve never used it.

Wanna do the 64 MB memory config cited in Wikipedia? I think 2 GiB is probably too much to expect for homelab people

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+- Inputs to the hash are:
+  - the plaintext password as UTF-8 bytes
+  - a random per-password salt (16 bytes)
+  - a random per-Wasmo instance **pepper** (TBD bytes - maybe >= 16 bytes to match salt?), passed as **associatedData** to the `Argon2` function.
```
swankjesse: Use the account ID also? I think that’s a good associated data

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+  - these hard coded values (from OWASP guidelines); the hard-coded values at the time a password is created are stored with the hashes, in case we want to change these values later.
+    - memory size m=12288 (12 MiB)
```
swankjesse, line 74: Rad

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+    - iterations t=3
+    - parallelism p=1
+- Hashing happens on the server only; during login, the client sends the plaintext password over https (else the hash would become the password); open question: Is this standard, or is there a better way?
```
swankjesse: This seems good to me. And if we can’t use HTTPS to send secrets then the cookie becomes the password.

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+- Protection against brute force cracking is initially only through the slowness of Argon2 + difficulty of obtaining DB + pepper. In the future, rate limiting will be added for all server calls (doesn't help protect leaked DB + pepper).
```
swankjesse: Please rate limit with `com.wasmo.permits.PermitService`

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+- We'll use the [reference implementation of Argon2](https://github.com/P-H-C/phc-winner-argon2) (in C) with these [JVM bindings](https://github.com/phxql/argon2-jvm) - that project is much more active than [the alternative](https://github.com/kosprov/jargon2-api) also linked from the reference implementation GitHub page.
```
swankjesse, line 82: Both look pretty reasonable to me. I’d love to see one that uses the new FFI.

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+- The per-Wasmo instance *pepper* value means that password hashes can only be used on that instance.
```
swankjesse, line 83: Interesting note: we will support moving *computers* between Wasmo instances, but not accounts. This is consistent with that plan.

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,34 @@ Passwords are optional.
...
+- The pepper value is secret configuration of the Wasmo instance, like the postgresql password. It's stored in the same place: For now in an environment variable, later in a dedicated secrets vault.
+- Out of the three variants (Argon2i, Argon2d, and Argon2id), we'll ues Argon2id for hybrid protection against both classes of attack noted in the [README](https://github.com/P-H-C/phc-winner-argon2/blob/master/README.md).
```
swankjesse, line 86: rad

swankjesse (approved):
Thanks for writing this up. I love it.

docs/code/accounts_and_authentication.md
15characterlimi: Where is your 2 GiB figure coming from?

The memory use for Argon2id will be much less. [OWASP's section on Argon2id](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id) proposes different options that are equivalent (trading off memory use vs iteration code and paralleism, to achieve a constant difficult).

I had arbitrarily picked the middle one of the five options given by OWASP (see lines 66-68 of this file), and that one uses 12 MiBytes of memory.

swankjesse: It's on the Wikipedia page. I like the config you chose.

15characterlimi: Ah, thanks, I had missed those on the Wikipedia page. Looks like they're from the original Argon2 RFC.

I generally trust OWASP so I expect that their lower numbers will be sufficient.

I plan to continue to follow OWASP recommendations, i.e. stick with the 12 MiBytes of memory, and resolve this thread.

docs/code/accounts_and_authentication.md
15characterlimi: Okay I'm changing this doc to associate the password with the account (potentially email based), not just with usernames.

But I'm still a bit uneasy about it.

For example, I'm guessing a rogue client shouldn't be able to probe whether a particular email address has an account - right? That would mean that the client can't find out whether the thing they're prompting the user to enter is a password vs. a challenge code.

How are we currently handling this for passkeys (I've never tried email address based/hosted accounts)? Can the user provide their passkey without the server given away that the account identified by the email address exists?

docs/code/accounts_and_authentication.md
15characterlimi: Ack, changing it to also support email address based accounts, although I'm still unconvinced that that is a good idea.

docs/code/accounts_and_authentication.md
15characterlimi: FYI this is obsolete now that the password is associated with the account rather than the (canonical) username, so this line no longer exists.

docs/code/accounts_and_authentication.md
15characterlimi: good point, tweaked the phrasing to note this.

docs/code/accounts_and_authentication.md
15characterlimi: Thanks, added this as a sub-bullet.

docs/code/accounts_and_authentication.md
15characterlimi: Added as an additional protection, along with a note that this layer of protection doesn't help once a dump of the DB has been stolen.

I'm not familiar with PermitsService so will  have to familiarize myself when I implement this plan.

docs/code/accounts_and_authentication.md
15characterlimi: Are you referring to java.lang.foreign (apparently added in Java 19)? TIL I learned that it exists :-)

15characterlimi (commented):
@swankjesse I addressed/responded to your comments and uploaded an "address review comments" commit. Mind having another look, since there were some details where our intuitions differed?

Sorry it took me so long to get back to you!

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,46 @@ Passwords are optional.
...
+- Inputs to the hash are:
+  - the plaintext password as UTF-8 bytes
+  - a random per-password salt (16 bytes)
+  - a random per-Wasmo instance **pepper** (TBD bytes - maybe >= 16 bytes to match salt?)
+  - the account ID
+  - these hard coded values (from OWASP guidelines); the hard-coded values at the time a password is created are stored with the hashes, in case we want to change these values later.
+    - memory size m=12288 (12 MiB)
+    - iterations t=3
+    - parallelism p=1
+- The pepper and accountId together are passed as **associatedData** in the `Argon2` function. While we could just concatenate them, we will instead serialize a protobuf containing those two values in order to remain more flexible about future additions.
```
swankjesse, line 77: I’m reluctant to introduce protobuf into this system. Perhaps 'newline-separated string' or JSON would be sufficient?

docs/code/accounts_and_authentication.md
```diff
@@ -44,7 +44,46 @@ Passwords are optional.
...
+- Hashing happens on the server only; during login, the client sends the plaintext password over https (else the hash would become the password); note that the cookie is transmitted over https as well.
+- Protection against brute force cracking is provided through
+   - slowness of Argon2 + difficulty of obtaining DB + pepper
+   - rate limiting via com.wasmo.permits.PermitService (this measure doesn't help if/when a dump of the DB has been stolen)
+- We'll use the [reference implementation of Argon2](https://github.com/P-H-C/phc-winner-argon2) (in C) with these [JVM bindings](https://github.com/phxql/argon2-jvm) - that project is much more active than [the alternative](https://github.com/kosprov/jargon2-api) also linked from the reference implementation GitHub page.
+- The per-Wasmo instance *pepper* value means that password hashes can only be used on that instance.
+  - This is consistent with our plan to support moving _computers_ between Wasmo instances, but not _accounts_.
```
swankjesse, line 84: YES!

swankjesse (commented):
LGTM

This is great!

docs/code/accounts_and_authentication.md
15characterlimi: JSON works, I'll probably go with that.

PR #148, opened 2026-08-19
Add support for passwords.

15characterlimi:
Add support for passwords.
 - DB storage of password hashes in `Password` table.
   - An empty password is the same as no password. It's currently represented as an absent row in the DB, but we could alternatively store the hash of "" - downside: We'd need to separately store the boolean for whether the password is empty, in order to not ask for one in the client UI.
   - Password hashes are computed via `Argon2PasswordHasher` from the plaintext password, accountId, per-password instance random salt, and per-wasmo instance pepper (read from environment variable).
 - `PasswordManager` builds on top of DB and `Argon2PasswordHasher` to provide a convenient API.
 - Rate limit is applied at the `UsernameRpcs` level. Since `PermitService.tryAcquire()` requires a `SqlConnection`, it happens within the same DB transaction that will then check the username/password (if the permit was successfully acquired).
 - Client-side UI for entering a password during Sign Up / Sign In
   - Sign Up: Ask for optional password (and, if given, a confirmation)
   - Sign In: Ask for password if the account has one (allows for password authentication)
   - Show an error message if the server rejects the sign-in (todo: make the error text red or similar).
   - wired up with corresponding code at the server-side RPC layer

TODOs:
 - Follow-up PR: UI for setting, clearing or changing a password once signed-in.
 - Follow-up PRs: security improvements:
    - encrypt/obfuscate the password with some secret shared between  client and server, so it doesn't appear in plaintext in any request logs.
    - potentially introduce a security policy, e.g. enforce password for new accounts or a minimum password length.
 - Password recovery or reset (e.g. via the admin account)
 
Open questions:
  - Is it okay to expose the list of available usernames, and whether they have a password, to signed-out clients? Note that de facto, usernams currently only exists on Homelab.
  - Should no-password be represented as (a) a missing row, or (b) a row with the salted and peppered hash of "", in the DB? Currently we do (a).
  
## Screenshots

image
image


docs/local_development/postgresql.md
```diff
@@ -1,6 +1,12 @@
 Local Development PostgreSQL
 ============================
 
+One-time setup (optionally okay to keep even when wiping the database):
+
+```bash
+$ echo 'export WASMO_PASSWORD_PEPPER="'$(dd if=/dev/urandom bs=1 count=16 2>/dev/null | base64 --wrap=0)'"' >> ~/.zshrc && source ~/.zshrc
```
swankjesse: Could we do this with schema migrations instead? Or is it useful to keep it independent from the database storage?

swankjesse: (Doesn’t matter much either way!)

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignInWithUsernameAndPasswordScreen.kt
```diff
@@ -0,0 +1,117 @@
...
+@Composable
+fun SignInWithUsernameAndPasswordScreen(
+  attrs: AttrsScope<HTMLDivElement>.() -> Unit = {},
+  eventListener: (SignUpEvent) -> Unit,
+  usernameOptions: List<UsernameSlug>,
+  passwordEntry: Pair<UsernameSlug, String>?,
+  canCreateUsername: Boolean,
+  canSubmit: Boolean,
+  busy: Boolean,
+  warningMessage: String? = null,
```
swankjesse: Is this a passwordWarningMessage? Pinning it to the password field if its present, and not if its not, suggests this might be two independent values?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
```diff
@@ -214,6 +267,27 @@ class SignUpPresenter(
     }
   }
 
+  // TODO: localize to the User's UI Locale
```
swankjesse: This TODO probably goes for the whole project!

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
```diff
@@ -188,17 +210,48 @@ class SignUpPresenter(
...
+          }
+        } else {
+          // clear any previous selection of a username and associated partially-typed password,
+          // then sign-in without password for the newly clicked username.
+          updateSignInWithUsernameAndPasswordModel { model ->
+            model.copy(
+              passwordEntry = null
+            )
+          }
+          attemptSignIn(usernameSlug, password = null) { linkUsernameDecision ->
```
swankjesse: Is the extra action always the same?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
```diff
@@ -188,17 +210,48 @@ class SignUpPresenter(
...
+            val errorMessage = linkUsernameDecision.toErrorMessage()
+            if (errorMessage != null) {
+              updateSignInWithUsernameAndPasswordModel { model ->
+                model.copy(
+                  warningMessage = errorMessage,
```
swankjesse: This uses the words 'warning' and 'error' as synonyms... I’d prefer either no prefix or to use the word 'error'? Because I believe this message indicates something the user must fix?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
```diff
@@ -214,6 +267,27 @@ class SignUpPresenter(
...
+  private fun attemptSignIn(usernameSlug: UsernameSlug, password: String?, linkUsernameDecisionExtraAction: ((LinkUsernameDecision) -> Unit)) =
+    callServer {
+      val response = accountDataService.linkUsername(usernameSlug, password ?: "")
+      if (response.decision == LinkUsernameDecision.Success) {
+        response.account?.let { accountSnapshot ->
+          accountDataService.receiveAccountSnapshot(accountSnapshot)
+        }
+        router.goTo(HomeRoute, TransitionDirection.PUSH)
+      }
+      linkUsernameDecisionExtraAction.invoke(response.decision)
```
swankjesse: Does this call belong in an 'else' clause? It makes me anxious to have function calls after the call to `router.goTo()`.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
```diff
@@ -134,6 +138,9 @@ class HomelabCommand : CliktCommand() {
           localPostgresql = localPostgresql,
           logger = logger,
           eventListener = eventListener,
+          hashingPepper = HashingPepper(
+              (System.getenv("WASMO_PASSWORD_PEPPER") ?: "").decodeBase64()!!
```
swankjesse: Perhaps best to crash if the pepper is absent. Otherwise when it is set later, it’ll break all of the passwords?

swankjesse: Perhaps for robustness + simplicity, we should move the pepper from being stored in an environment variable to being stored in the DB? I think that reduces setup – all state is in the database!

os/server/db/src/jvmMain/kotlin/com/wasmo/db/passwords/DbPassword.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.wasmo.identifiers.PasswordHash
+import com.wasmo.identifiers.PasswordId
+import kotlin.time.Instant
+import okio.ByteString
+
+data class DbPassword(
+  val id: PasswordId,
+  val createdAt: Instant,
+  val accountId: AccountId,
+  val passwordHash: PasswordHash, // salted and peppered
```
swankjesse: The ‘salted and peppered‘ comment applies to the type `PasswordHash` and not this specific field, right?

os/server/db/src/jvmMain/resources/migrations/v5__PasswordDigest.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE Password (
+  id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  account_id BIGINT NOT NULL UNIQUE,
+  -- Subject to clock limitations (e.g. may not be strictly increasing).
```
swankjesse, line 4: This is true of all timestamps in our system, right? I don’t disagree with the comment but noting it here suggests it’s the password field’s created_at that has this limitation, and not the whole DB

os/server/db/src/jvmMain/resources/migrations/v5__Password.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE Password (
+  id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  account_id BIGINT NOT NULL UNIQUE,
+  -- Subject to clock limitations (e.g. may not be strictly increasing).
+  created_at TIMESTAMPTZ NOT NULL,
+  -- Includes metadata, most importantly the salt (but not the pepper).
```
swankjesse: Whoa, hang on. This ‘not the pepper' is not consistent with the PasswordHash field above. That is suspect!

os/server/db/src/jvmMain/resources/migrations/v5__Password.sql
```diff
@@ -0,0 +1,8 @@
...
+  password_hash TEXT NOT NULL
```
swankjesse: Oooh I think this should be `BYTEA` not `TEXT`. I don’t want to have to guess whether it’s hex or base64 at query time!

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
```diff
@@ -1,11 +1,47 @@
...
 
+import java.security.MessageDigest
 import kotlinx.serialization.Serializable
+import okio.ByteString
 
 @Serializable
 @JvmInline
 value class AccountId(val id: Long)
 
+// avoid putting com.wasmo.identifiers.Secure into our API signature
```
swankjesse, line 14: Why not?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
```diff
@@ -1,11 +1,47 @@
...
+  //    it's necessary for it to run in constant time to twart timing based oracle attacks.
+  fun secretValueChars() = secretValue.toCharArray()
+
+  fun isEmpty() = secretValue.isEmpty()
+
+  fun isNotEmpty() = secretValue.isNotEmpty()
+  override fun toString(): String = "<redacted password>"
+
+  companion object {
+    val EMPTY = AccountPassword("")
```
swankjesse, line 36: I’m not particularly convinced on the not overriding equals thing? Might be simpler to implement a basic time-constant equals function? Otherwise it’s a bit surprising that `password.isEmpty()` is not consistent with `password == Password.EMPTY`. (And I’d certainly prefer if comparing against `EMPTY` was sufficient!)

You can do a time-resistant equals like so:
```
return secretValue.encodeUtf8().sha256() == other.secretValue.encodeUtf8().sha256()
```

(There’s probably more efficient strategies, but this one is easy to reason about)

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
```diff
@@ -1,11 +1,47 @@
...
+  }
+}
+
+@JvmInline
+value class HashingPepper(val value: ByteString) {
+  init {
+      require(value.size == 16) {
+        // In the future we may want to allow >= 16 bytes
```
swankjesse: I don’t think this comment is necessary. (All code is subject to change!)

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
```diff
@@ -1,11 +1,47 @@
...
+        "Hashing pepper should be exactly 16 bytes."
+      }
+  }
+}
+
+// Note: Two PasswordHash instances generated for the same password / accountId are not in general
+// equal, because they also depend on a random salt (and also a per-instance pepper).
```
swankjesse: _instance_ is a bit of a tricky word in this sentence. per-service pepper?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
```diff
@@ -1,11 +1,47 @@
...
+@JvmInline
+value class PasswordHash(val value: String)
```
swankjesse: I see the word 'hash' and the type 'string' and something doesn’t seem right

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordHasher.kt
```diff
@@ -0,0 +1,31 @@
...
+   * different (modulo the infinitesimal chance of generating the same salt twice) for successive
+   * invocations with the same parameter values.
+   *
+   * Implementations may in addition choose to pepper the value to be hashed, in which case the same
+   * pepper must be configured for a hashed value to later [verify].
+   */
+  fun hash(password: AccountPassword, accountId: AccountId): PasswordHash
+
+  /**
+   * Checks whether the given password matches the given hash. Implementation who pepper the
```
swankjesse: This says 'implementation who pepper' but that’s all implementations, right? We’re not shipping a non-peppering system

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordManager.kt
```diff
@@ -0,0 +1,12 @@
+package com.wasmo.passwords
+
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.AccountPassword
+import wasmox.sql.SqlTransaction
+
+interface PasswordManager {
+  context(sqlTransaction: SqlTransaction)
+  suspend fun isPasswordCorrect(accountId: AccountId, providedPassword: AccountPassword): Boolean
```
swankjesse: ooooh I don’t think I like the signature of this function because it reads like a side-effect free function, and I think it absolutely should have a side-effect (incrementing a rate limit).

Perhaps `checkPassword()` ?

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordManager.kt
```diff
@@ -0,0 +1,12 @@
+package com.wasmo.passwords
+
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.AccountPassword
+import wasmox.sql.SqlTransaction
+
+interface PasswordManager {
```
swankjesse: As a rule I try to avoid the word 'Manager' because there's almost always a more semantic word available instead. PasswordAuthenticator?

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,94 @@
...
+import com.wasmo.identifiers.OsScope
+import com.wasmo.identifiers.PasswordHash
+import de.mkammerer.argon2.Argon2Factory
+import dev.zacsweers.metro.Inject
+import dev.zacsweers.metro.SingleIn
+
+
+/**
+ * For values returned by [hash] to later [verify], the same [pepper] value must be specified
+ * on the respective [Argon2PasswordHasher] instances.
```
swankjesse: Hell yeah

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,94 @@
...
+    const val PARALLELISM = 1
+
+    val TYPE = Argon2Factory.Argon2Types.ARGON2id
+    val PASSWORD_CHARSET = Charsets.UTF_8
+  }
+
+  // Use default salt and hash length (16 and 32 bytes, respectively, as of Argon2 library v2.12).
+  private val argon2 = Argon2Factory.create(Argon2Config.TYPE)
+
+  private fun Long.toByteArrayLittleEndian() = ByteArray(size = Long.SIZE_BYTES)
```
swankjesse: Alternately, `Buffer().writeLongLe(this).readByteArray()`

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,94 @@
...
+    try {
+      dataElements += password.secretValue.toByteArray(Argon2Config.PASSWORD_CHARSET)
+      dataElements += pepper.value.toByteArray()
+      dataElements += accountId.id.toByteArrayLittleEndian()
+      return dataElements.concat()
+    } finally {
+      // TODO: Maybe drop wipeArray()? - kind of pointless when password: String is on the heap.
+      dataElements.forEach { argon2.wipeArray(it) }
+    }
+  }
```
swankjesse:
```suggestion
  private fun dataBytes(
    password: AccountPassword,
    accountId: AccountId,
  ): ByteArray {
    val buffer = Buffer()
    buffer.writeUtf8(password.secretValue)
    buffer.write(pepper.value)
    buffer.writeLongLe(accountId)
    return buffer.readUtf8()
  }
```

docs/local_development/postgresql.md
15characterlimi: A pepper that's stored in the DB would be pointless because then it'd effectively be part of the salt.

As I understand it, the pepper's purpose is to ensure that even if the hashing algorithm can be inverted and even if the database leaks, the passports remain secure. That's why it needs to be stored outside of the DB.

The pepper also has the (neutral, I think) side effect that the passwords in DB snapshots can't be used by another instance (which uses a different pepper).

That said, I'm open to storing the pepper elsewhere that's at arm's length from the database. Environment variables were the best idea I had, but if you have a better idea, I'm open to it!

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignInWithUsernameAndPasswordScreen.kt
15characterlimi: It always relates to a username (that we tried to sign-in as or perhaps tried to create) but not always to a password. It's the error message returned by the server side linkUsername or createAccount RPC endpoints .

The reason it's sometimes below the list of usernames is twofold:

 - On sign-in attempt for a username that doesn't have a password, the model doesn't remember which username was clicked - we just directly try the sign-in when the username is clicked.
 - For sign-up, the username we tried to create isn't in the list.

I could

(a)
- change the model to remember what username was clicked
- create a data class UsernameRelatedError(username: UsernameSlug, errorMessage): String
- replace warningMessage: String? with usernameRelatedError: UsernameRelatedError?

and then still show the error message either in-line in the list of usernames, or else below. Or I could 

(b)
have two separate 

  signInError: UsernameRelatedError?
  signUpError: UsernameRelatedError?

or

(c) just rename warningMessage -> errorMessage, perhaps with a comment on top.

I'm leaning towards (c) for simpicity. WDYT?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
15characterlimi: dropping the TODO

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
15characterlimi: You're right, it is. I'll deduplicate the code and fold it into the `callServer { ... }` block inside `attemptSignIn()`, unless you'd prefer me to change `callServer` to return a `Deferred<T>` and do the model update outside of `attemptSignIn`?

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
15characterlimi: Changing to errorMessage, since it's a (linkAccount or createAccount) error returned by the server.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
15characterlimi: Moved the whole errorMessage-related block to above the handling of the Success case.

Yes the errorMessage will always be`null` in case of `Success`, but I'd very slightly prefer the client to not hard-code / duplicate this assumption.

I don't feel strongly though, happy to change,

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
15characterlimi: 1.) It does crash if the pepper env var is absent, because HashingPepper ctor enforces that the value is exactly 16 bytes long.

Given that this was nonobvious, I've wrapped the System.getEnv() call in requireNotNull(), although the fact that getenv() will always return null and isn't free to instead return "" for undefined env variables is still a tiny bit obscure.

---

2.) On your second comment: The pepper **must**  live outside the DB, else it is the salt. The use of a pepper as an additional defense in depth is recommended by https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html but not mandatory.

In practice, other parts of our system (that passwords are sent to our RPC endpoints as plain strings, although over https) is probably a bigger attack surface than if we had unpeppered password hashes, but I hope to fix this in a follow-up PR.

Do you think peppering's security benefits pull their weight to warrant the additional project setup step? 

I considered having the server generate a pepper and write it to a file on first launch, but I figured that would not really add any security because that file is more likely to leak than an env variable when the DB leaks. WDYT?

os/server/db/src/jvmMain/kotlin/com/wasmo/db/passwords/DbPassword.kt
15characterlimi: It applies to the implementation `Argon2PasswordHasher` that we use to produce the PasswordHash instances - it's not inherent in either the PasswordHash type nor in the DB column.

The comment is a bit fragile in that we'll forget to update the comment if we ever change the behavior, OTOH it might be useful information. For now I'm changing the comment to:

// Argon2PasswordHasher produces passwordHashes that are salted and peppered

os/server/db/src/jvmMain/resources/migrations/v5__PasswordDigest.sql
15characterlimi: Yeah I believe it applies to all timestamps in our system, as long as they derive from the real clock which has this characterstics - other Clock implementations do not.

I put it here to remind a reader in case they were going to sort passwords by timestamp.

I'll resolve this comment thread, but if there's any action you'd like me to take then please feel free to reopen it!

os/server/db/src/jvmMain/resources/migrations/v5__Password.sql
15characterlimi: I think it is inconsistent?

The PasswordHash is peppered (as long as it's produced by Argon2PasswordHasher, which takes a HashingPepper ctor argument).

That doesn't mean that the pepper should be stored with the password.

We specifically choose to store the salt with the password (in the DB) while the pepper lives elsewhere, due to the implied security properties.

WDYT? Perhaps I need to tweak some comments somewhere to make this clearer?

os/server/db/src/jvmMain/resources/migrations/v5__Password.sql
15characterlimi: The Argon2 .java wrapper library returns a String, though, not a ByteArray. This is because it's really a composite value containing both the raw hash as well as associated metadata (salt, version of Argon2i, and perhaps other config values).

BouncyCastle does return a ByteArray, but in my abandoned draft implementation of PasswordHasher on top of BouncyCastle, I constructed a composite JSON String that contained the base64-encoded raw hash along with the salt.

So I think that String / TEXT is the correct type for this. Perhaps the word password_hash is wrong, though - since you apparently thought it was going to be the raw hash? Would it be clearer if we renamed PasswordHash -> PasswordDigest, password_hash -> password_digest or something?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
15characterlimi: Chiefly because it's from a different gradle module (although the same packageName com.wasmo.identifiers).

AFAIK, that other module is shared between client and server, whereas this one here is specific to the server. Perhaps the packageNames should be distinct and have "server" vs.  { "", "shared" or similar } in them? WDYT?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
15characterlimi: removed

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
15characterlimi: As noted above, how about Digest?

This isn't the raw hash, it's a composite value including the hash, salt, and any other metadata (e.g. version of Argon2i), the way it is encoded to a String is an implementation detail of the library.

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordHasher.kt
15characterlimi: That's correct because the only implementation of PasswordHasher right now is Argon2PasswordHasher

Would you prefer if I turned PasswordHasher into an abstract class, so that hashingPepper can move up into its ctor?

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordManager.kt
15characterlimi: PasswordAuthenticator to me sounds like it only checks passwords but doesn't also set / persist them. Given that this abstraction does both, do you still prefer PasswordAuthenticator over PasswordManager?

15characterlimi: Brainstorming other names (I think they're worse): PasswordService, PasswordChecker, PasswordOperations

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
15characterlimi: no longer needed now that we have the new implementation of dataBytes()

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordManager.kt
15characterlimi: Renamed to checkPassword(), although that makes the semantics of true / false return values less obvious.

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
15characterlimi: How about "and also on the pepper configured on the PasswordHasher"?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
15characterlimi: sha256() probably still depends on the length of secretValue, which is okay but then we may as well use java.security.MessageDigest.isEqual(ByteArray, ByteArray).

Implemented equals and hashCode now, on that basis.

docs/local_development/postgresql.md
swankjesse: Sounds good. Our options are environment variables, and the `/wasmo` directory in homelab.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignInWithUsernameAndPasswordScreen.kt
swankjesse: I think definitely C.

The other option I consider is having a `usernameMessage` and a `passwordMessage`, rather than moving the message around depending on what fields are present.

...That leads to this, which I’m also okay with:

```
class FieldMessage(
  val location: FieldId,
  val message: String,
)
```

But that’s far into follow-up territory.

I believe our CSS framework has Nice UI for these

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
swankjesse: No preference there.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
swankjesse: I want the homelab setup to be EXTREMELY QUICK and easy. The biggest drawback of the pepper I see is that it’s fragile . . . as with all environment variables you could lose it if you change which UNIX user your process runs as, or whatever. And unlike passwords and database names, if you lose the pepper it just makes the system seem broken.

os/server/db/src/jvmMain/kotlin/com/wasmo/db/passwords/DbPassword.kt
swankjesse: Lets assume we only have one non-testing PasswordHasher? There’s diminishing utility to making such things configurable!

os/server/db/src/jvmMain/resources/migrations/v5__PasswordDigest.sql
swankjesse: I think it’s excellent to remind engineers that clocks might not be monotonic, but I think that putting this comment here on this line suggests that it’s special in a way that our other created_at dates are not.

os/server/db/src/jvmMain/resources/migrations/v5__Password.sql
swankjesse: Make it a 2-field thing? I dislike the idea that changing our Argon2 library would change how the password hashes are represented?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
swankjesse: Yeah, I think that’s right. The current `com.wasmo.identifiers` name is ambiguous. Follow-up!

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
swankjesse: Sounds good!

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
swankjesse: `PasswordArgon2` ? Being concrete here seems useful, and it reminds us that we can’t swap this out for another algorithm

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordHasher.kt
swankjesse: Lets lock it in right now - we will not support a different implementation of PasswordHasher. We have one that’s sufficient, and supporting alternatives would introduce unnecessary complexity.

(I don’t want our code to try to support alternate implementations that we’re unlikely to need.)

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordManager.kt
swankjesse: `PasswordService` or `PasswordStore` ?

docs/local_development/postgresql.md
15characterlimi: Is the /wasmo directory where the DB is also stored, or is it elsewhere because postgresql is a separate service?

Storing the pepper in a different directory tree from the DB would potentially be a good solution IMO, if it means that it frees us from the project setup step of having to set the env variable; it also means we can guarantee that the value is randomly generated (as opposed to every user copying the same hard coded value to their env variables).

Otherwise, let's stick with the env variable.

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignInWithUsernameAndPasswordScreen.kt
15characterlimi: Okay, I'll stick with C for now them - the "address review comments" commit has already done this, so I'll resolve this comment thread.

I had also thought of the class FieldMessage approach as a way to take option B, but I'm not convinced that it currently pulls its weight at this point.

Resolving this comment thread, since I think it's been sufficiently addressed..

os/client/app/src/jsMain/kotlin/com/wasmo/client/app/signup/SignUpPresenter.kt
15characterlimi: Taking no further action and resolving this comment thread.

In the "address review comments" commit, I had folded the logic into attemptSignIn():

```
private fun attemptSignIn(usernameSlug: UsernameSlug, password: String?) =
    callServer {
      val response = accountDataService.linkUsername(usernameSlug, password ?: "")
      val errorMessage = response.decision.toErrorMessage()
      if (errorMessage != null) {
        updateSignInWithUsernameAndPasswordModel { model ->
          model.copy(
            errorMessage = errorMessage,
          )
        }
      }
      if (response.decision == LinkUsernameDecision.Success) {
        response.account?.let { accountSnapshot ->
          accountDataService.receiveAccountSnapshot(accountSnapshot)
        }
        router.goTo(HomeRoute, TransitionDirection.PUSH)
      }
    }
```

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordManager.kt
15characterlimi: Personally I like both of those less than PasswordManager, but both are acceptable. Do you have a preference? If not, I'll go with PasswordStore, since the methods have `context(sqlTransaction: SqlTransaction)`.

15characterlimi: Renamed to PasswordStore. Resolving this comment thread.

15characterlimi:
@swankjesse Heads up, I've now also made the following changes:

  - Renamed Password -> PasswordDigest in a few places (PasswordHasher and HashingPepper keep their names for now)
  - Moved pepper: HashingPepper from the Argon2PasswordHasher ctor into the PasswordHasher base class ctor. I considered also moving the dataBytes() function up into the base class and potentially even making PasswordHasher a final class parameterized via two new ctor arguments `hash: ((ByteArray) -> PasswordDigest)` and `verify: ((ByteArray) -> Boolean)`, but have not done this for now.
  
  I've kept these changes to a separate commit for now, but plan to squash everything into the base commit before merging.

docs/local_development/postgresql.md
swankjesse: Env variable works for me. 

We should figure out exactly what we want for the pepper in Homelab. It’s currently nicely self-contained with zero external environment variables!

swankjesse: Lets use hex instead of base64. Base64 is awkward with its trailing `=` characters and its case-sensitivity!

```suggestion
$ echo 'export WASMO_PASSWORD_PEPPER="$(xxd -l 16 -p /dev/random)"' >> ~/.zshrc && source ~/.zshrc
```

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
```diff
@@ -134,6 +138,11 @@ class HomelabCommand : CliktCommand() {
           localPostgresql = localPostgresql,
           logger = logger,
           eventListener = eventListener,
+          hashingPepper = HashingPepper(
+              requireNotNull(
+                System.getenv("WASMO_PASSWORD_PEPPER")
+              ).decodeBase64()!!
+          )
```
swankjesse, line 146:
```suggestion
          hashingPepper = HashingPepper(
              requireNotNull(
                System.getenv("WASMO_PASSWORD_PEPPER")
              ).decodeHex()
          )
```

os/server/db/src/jvmMain/kotlin/com/wasmo/db/passwords/DbPasswordDigest.kt
```diff
@@ -0,0 +1,13 @@
+package com.wasmo.db.passwords
+
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.PasswordDigest
+import com.wasmo.identifiers.PasswordDigestId
+import kotlin.time.Instant
+
+data class DbPasswordDigest(
```
swankjesse, line 8: Nice

os/server/db/src/jvmMain/kotlin/com/wasmo/db/passwords/DbPasswordDigest.kt
```diff
@@ -0,0 +1,13 @@
...
+  val id: PasswordDigestId,
+  val createdAt: Instant,
+  val accountId: AccountId,
+  val passwordDigest: PasswordDigest, // Argon2PasswordHasher produces passwordHashes that are salted and peppered
```
swankjesse: Move the comment to `PasswordDigest` ?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
```diff
@@ -1,11 +1,57 @@
...
+        "Hashing pepper should be exactly 16 bytes."
+      }
+  }
+}
+
+/**
+ * A composite value composing salted hash of the password, the salt, and metadata (e.g. hash
+ * algorithm / version used). Because the hashes are salted and peppered, digests of the same
+ * password will in general not be equal.
+ */
```
swankjesse, line 56: This comment is perfect

That said, it does make me want to switch to BouncyCastle (or another way to do Argon2) to avoid having a compound value, where the mechanism to encode/decode is an implementation detail of our Java-Argon2 binding. 

(I’m imagining some future person trying to switch us to a different Argon2 library and having to reproduce this encoding)

os/server/db/src/jvmMain/resources/migrations/v5__PasswordDigest.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE PasswordDigest (
+  id BIGINT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+  account_id BIGINT NOT NULL UNIQUE,
+  -- Subject to clock limitations (e.g. may not be strictly increasing).
+  created_at TIMESTAMPTZ NOT NULL,
+  -- Includes raw hash, salt (but not pepper), and metadata such as hash algorithm version used.
```
swankjesse: This doc is repetitive with the (excellent) docs on PasswordDigest type. Omit this?

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordHasher.kt
```diff
@@ -0,0 +1,34 @@
...
+   * hasing algorithm version used). The structure of the returned String is an implementation
+   * detail which clients should treat as opaque.
+   *
+   * Implementation must securely generate a new random salt on each invocation, and use that
+   * salt as well as [pepper] as an input to the hash.
+   *
+   * Therefore, for values returned by [digest] to later [verify], the same [pepper] value must
+   * be specified at construction time and the same PasswordHasher implementation, or a backwards
+   * compatible later version, should be used.
+   */
```
swankjesse, line 24: Perfect docs, no notes

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordStore.kt
```diff
@@ -0,0 +1,15 @@
+package com.wasmo.passwords
+
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.AccountPassword
+import wasmox.sql.SqlTransaction
+
+interface PasswordStore {
```
swankjesse, line 7: Does this enforce rate limits? We should document that this enforces rate limits.

(I think this is probably the best place to enforce rate limits, because otherwise all callers need to do that)

swankjesse: (And I think we might want rate limits on both checking and setting? That might be something OWASP has guidance for?)

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordStore.kt
```diff
@@ -0,0 +1,15 @@
...
+  /**
+   * Returns true of the provided password matches the password recorded for the given AccountId.
+   */
+  context(sqlTransaction: SqlTransaction)
+  suspend fun checkPassword(accountId: AccountId, providedPassword: AccountPassword): Boolean
```
swankjesse, line 12: Should this return GoodPassword / BadPassword / RateLimited ?

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,77 @@
...
+import okio.Buffer
+
+
+/**
+ * Produces salted, peppered [PasswordDigest]s containing metadata about the hashing algorithm
+ * configuration, the raw hash and the salt.
+ *
+ * Example value produced:
+ *
+ * `"$argon2id$v=19$m=12288,t=3,p=1$jgHWZdnPF/Qg63L4FxeA7Q$MenB3YE56XIKXdICSAnzQznH+CAbe/8jF6u4Vh79Fio"`
```
swankjesse: Either now or in follow-up, I think we split this from a compound string into a data class

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,77 @@
...
+ */
+@Inject
+@SingleIn(OsScope::class)
+class Argon2PasswordHasher(pepper: HashingPepper): PasswordHasher(pepper) {
+
+  // These parameters were decided in docs/code/accounts_and_authentication.md
```
swankjesse: Nice docs.
```suggestion
  /** These parameters were decided in docs/code/accounts_and_authentication.md */
```

The benefit of `/**` comments over `//` comments is that the IDE shows the `/**` comments at callsites with CTRL+J. Double-tap CTRL+J for a persistent docs pane!

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,77 @@
...
+  object Argon2Config {
+    const val MEMORY = 12288
+    const val ITERATIONS = 3
+    const val PARALLELISM = 1
+
+    val TYPE = Argon2Factory.Argon2Types.ARGON2id
+  }
+
+  // Use default salt and hash length (16 and 32 bytes, respectively, as of Argon2 library v2.12).
```
swankjesse: Should we hardcode these, so library updates can’t weaken/strengthen this config?

swankjesse: One software update cost one company’s customers $140M because it weakened a random number generator! Explicit defaults are good!

https://blog.coinkite.com/entropy-technical-backgrounder/

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,77 @@
...
+  ): PasswordDigest {
+      val data: ByteArray = dataBytes(password, accountId)
+      try {
+        // The .java wrapper of the argon2 library internally generates a new salt each time this
+        // function is called; the value of that salt cannot be customized via the API surface.
+        // That salt value is embedded within the returned hashString.
+        val hashString = argon2.hash(Argon2Config.ITERATIONS, Argon2Config.MEMORY, Argon2Config.PARALLELISM, data)
+        return PasswordDigest(hashString)
+      } finally {
+        argon2.wipeArray(data)
```
swankjesse: probably not worth the effort, as AccountPassword.secretValue is a String

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/RealPasswordStore.kt
```diff
@@ -0,0 +1,69 @@
...
+import com.wasmo.db.passwords.findPasswordDigestForAccount
+import com.wasmo.db.passwords.insertPasswordDigest
+import com.wasmo.db.passwords.removePasswordDigest
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.AccountPassword
+import kotlin.time.Clock
+import wasmox.sql.SqlTransaction
+
+/**
+ * Note: This class doesn't apply rate limiting. Rate limiting should be applied at the RPC layer.
```
swankjesse: Ooooh this doc probably belongs on the interface?

I would like to move the rate limiting to here though. Otherwise we have an API that does something dangerous (checking a password without a rate limit), and its on the callers to be judicious

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/RealPasswordStore.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    return if (providedPassword.isNotEmpty() && dbPasswordDigest != null) {
+      passwordHasher.verify(
+        password = providedPassword,
+        accountId = accountId,
+        passwordDigest = dbPasswordDigest.passwordDigest,
+      )
+    } else {
+      val isPasswordCorrect = (providedPassword.isEmpty() && dbPasswordDigest == null)
+      // waste some cycles to help thwart timing based oracle attacks trying to figure out if an account exists.
```
swankjesse, line 39: This is so cool

os/server/passwords/real/src/jvmTest/kotlin/com/wasmo/passwords/Argon2PasswordHasherTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import assertk.assertions.isFalse
+import assertk.assertions.isTrue
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.AccountPassword
+import com.wasmo.identifiers.HashingPepper
+import okio.ByteString.Companion.encodeUtf8
+import org.junit.Test
+
+class Argon2PasswordHasherTest {
+  val pepper = HashingPepper("pepper for tests".encodeUtf8()) // 16 bytes
```
swankjesse, line 18: neato sample pepper

os/server/passwords/real/src/jvmTest/kotlin/com/wasmo/passwords/Argon2PasswordHasherTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import assertk.assertThat
+import assertk.assertions.isFalse
+import assertk.assertions.isTrue
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.AccountPassword
+import com.wasmo.identifiers.HashingPepper
+import okio.ByteString.Companion.encodeUtf8
+import org.junit.Test
+
+class Argon2PasswordHasherTest {
```
swankjesse, line 17: What a perfect test

os/server/permits/api/src/jvmMain/kotlin/com/wasmo/permits/PermitService.kt
```diff
@@ -6,6 +6,8 @@ import kotlin.time.Instant
 import wasmo.sql.SqlConnection
 
 interface PermitService {
+  // TODO: Make this return a Permit? so we can permit.release() rather than having to repeat
+  // (now, type, value, -count) to release acquired permits back into the pool.
```
swankjesse, line 10: 👍🏻

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
```diff
@@ -14,33 +14,47 @@ import com.wasmo.db.usernames.selectLinkedUsernameOrNullAllowDeleted
...
 import dev.zacsweers.metro.Inject
 import dev.zacsweers.metro.binding
 import kotlin.time.Clock
+import kotlin.time.Duration.Companion.hours
 import org.jetbrains.annotations.VisibleForTesting
 import wasmo.sql.SqlDatabase
 import wasmox.sql.SqlTransaction
 import wasmox.sql.transaction
 
+val PasswordAuthenticationRateLimit = RateLimit(
```
swankjesse, line 35: This seems a little generous? I’m imagining my kid breaking into my homelab secured by my dog’s name.

Changing it to 5 attempts per 5 minutes means you can’t blitz through 50 possibilities immediately; you gotta slow down and spend 45 minutes

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
```diff
@@ -50,17 +64,41 @@ class LinkUsernameRpc(
...
+    ) {
+      // For username based sign-in specifically, the existence of the account is not secret,
+      // so we can leak the fact that the account exists but password authentication failed.
+      LinkUsernameDecision.PasswordAuthenticationFailed
     } else {
       client.signIn(
         sourceAccountId = client.getOrCreateAccountId(),
         targetAccountId = existingUsername.accountId,
       )
+      // On success, release the permit back into the pool rather than consuming it.
```
swankjesse, line 96: nice

os/server/usernames/real/src/jvmTest/kotlin/com/wasmo/usernames/UsernameRpcsTest.kt
```diff
@@ -76,7 +81,71 @@ class UsernameRpcsTest {
...
+  fun `create existing username fails`() = runTest {
+    val client = tester.newClient()
+    val username = client.createNewUsername()
+    val createResponseA = client.createUsername(username = username, password = "password")
+    val createResponseB = client.createUsername(username = username, password = "password")
+    val createResponseC = client.createUsername(username = username, password = "wrong password")
+
+    val decisions = listOf(createResponseA, createResponseB, createResponseC).map { it.body.decision }
+
+    assertThat(decisions).isEqualTo(listOf(
```
swankjesse, line 116: so rad

swankjesse (approved):
# 🚀🚀🚀🚀🚀

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordStore.kt
swankjesse: (follow-up is fine)

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/RealPasswordStore.kt
15characterlimi: Note that this is forward-looking: It will only provide security for when we offer to secure non-username based accounts (e.g. by email) to be secured by password.

Right now only username based accounts are wired up to be protected by passwords, and the list of usernames for which an account exists is public information (is handed out to a client that is not even signed in).

os/server/passwords/real/src/jvmTest/kotlin/com/wasmo/passwords/Argon2PasswordHasherTest.kt
15characterlimi: I had actually forgotten an assertion, in fun `successive hashes differ even for the same parameters and same pepper`. Added now :-)

os/server/usernames/real/src/jvmMain/kotlin/com/wasmo/usernames/UsernameRpcs.kt
15characterlimi: Changed to 5 attempts per 5 minutes. That's probably fine for homelab, but once we support password authentication potentially on other distributions that may have lots of accounts, we may have to revisit.

15characterlimi (commented):
I still have more comments to address when I have more time, here are responses to just some of them.

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
15characterlimi: Okay, I will look into switching to BouncyCastle (I already have a WIP change which I shelved after you kindly pushed the new docker images with argon2-lib bundled). Will probably do it as part of this same PR. Stay tuned.

Leaving this comment thread open.

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
15characterlimi: done

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
15characterlimi: I agree (I had a comment somewhere about how the wipeArray() stuff is pointless). Will remove them all later today, stay tuned.

docs/local_development/postgresql.md
15characterlimi: On my Ubuntu at least I need a '  ... ' there, so went with the following:

`echo 'export WASMO_PASSWORD_PEPPER="'$(xxd -l 16 -p /dev/random)'"'  >> ~/.zshrc && source ~/.zshrc`

Hope that works on OS X!

Of course, I use .bashrc but since y'all use zsh on OS X, I stuck with .zshrc.

15characterlimi: P.S. Thanks for teaching me something new:

I saw you also replaced /dev/urandom with /dev/random, thanks for that! For some reason in my mind I thought /dev/urandom was the one that was guaranteed to be based on true randomness from hardware while /dev/random might be pseudorandom. Turns out it's the other way around! #TIL

Going with /dev/random, in line with your suggestion.

15characterlimi: Ah, I was assuming that the postgres auth credentials would be in env variables but turns out it's currently hard coded - I think that's probably worse?

If you think it's better then potentially we could ask the Homelab user to hard code the pepper there too, but that can easily lead to multiple instances running with the same pepper - which is bad, but so is multiple instances running with the same PostgreSQL auth credentials.

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
15characterlimi: Adopted manually!

So odd that decodeHex() throws while decodeBase64 returns null...

os/server/db/src/jvmMain/resources/migrations/v5__PasswordDigest.sql
15characterlimi: dropped the comment

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordStore.kt
15characterlimi: This does not currently implement rate limits, see this comment at the top of RealPasswordStore.kt:

**Note: This class doesn't apply rate limiting. Rate limiting should be applied at the RPC layer.**

Do you want the rate limits to be applied both at the RPC level and here or only here?

For now I'm replacing the commend in RealPasswordStore.kt with:

**TODO: Apply rate limiting here (and document this at the PasswordStore interface, not at the RPC level.**

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordStore.kt
15characterlimi: Only if/when RealPasswordStore implements rate limiting. I've added this line to the toplevel comment in RealPasswordStore.kt for now:

**This requires the PasswordStore methods to indicate whether a call was rate limited (exception or different return value).**

docs/local_development/postgresql.md
swankjesse: Upon closer inspection, we actually want `urandom` everywhere, including here. It doesn’t matter in this specific instance, but it’d be a good habit to prefer urandom.

https://www.thomas-huehn.com/myths-about-urandom/

swankjesse: I’d like to make it extremely easy to install Wasmo in a homelab, ideally a Docker image with a single writable directory mounted.

https://github.com/wasmcomputercompany/wasmo/blob/main/docs/distributions/homelab.md#run

This means that if you can access the file system of the computer that hosts Wasmo, you can capture its data. But there’s no other place for us to store secrets - that file system also hosts the env variables.

(We could potentially try to hook into the host operating system’s secure enclave?)

os/distributions/homelab/src/main/kotlin/com/wasmo/distributions/homelab/HomelabCommand.kt
swankjesse: Not odd. Just a giant mistake that I make. I have regrets!

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordStore.kt
swankjesse: Perfect

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordStore.kt
swankjesse: Sounds good

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
15characterlimi: Done: I pushed a new commit "Switch Argon2PasswordHasher implementation to BouncyCastle." that switches us to BouncyCastle.

I've updated Argon2PAsswordHasherTest.kt and the example value at the toplevel kdoc of Argon2PasswordHasher accordingly.

PTAL.

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
15characterlimi: Done in my "Switch Argon2PasswordHasher implementation to BouncyCastle." commit (pushed just now). PasswordDigest is that composite data class, and we store it in the DB using JSON serialization.

It would have been possible with the library I was previously using, because that one only gave us the fully-formed serialized digest string, without us having any control over its format.

PTAL.

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
15characterlimi: Done now as part of my "Switch Argon2PasswordHasher implementation to BouncyCastle." commit. All config values live inside Argon2PasswordHasher.kt, in the two fields `libraryConfig` and `wasmoConfig`.

These are currently _not_ ctor parameters because we don't need this (not even for tests), but it's possible that we'll to make them configurable via ctor parameters in the future in order to support tests for backwards compatibility behavior.

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/RealPasswordStore.kt
15characterlimi: Left it here for now since I'm leaning towards the responsibility change (move rate limiting here) for a follow-up PR, unless you feel strongly?

os/server/usernames/real/src/jvmTest/kotlin/com/wasmo/usernames/UsernameRpcsTest.kt
15characterlimi: I thought I already responded to this but looks like it got lost?

Since you liked this style, I've also tweaked the previous test method (`create and then link username enforces password`) to follow a different style, _but_ that one asserts a `map` rather than a `list`.

Pro: Don't have to mentally map the order
Con: Repeats the keys of the list

Do you have a preference?

docs/local_development/postgresql.md
15characterlimi: Intuitively it seems to me that capturing (stealing) the file system content will be easier for somehow who steals a copy of the DB than stealing the environment variables would be, but I don't have empirical data on how much each approach thwarts attackers.


Suggestion:

- Denote a local file ~/.wasmo/pepper.dat or similar (thoughts on location?)
- At startup, consult WASMO_PASSWORD_PEPPER environment variable
 - If the env variable doesn't exist, fall back to reading from a local file, ~/.wasmo/pepper.dat or similar (thoughts on location?)
 - If the file also doesn't exist, attempt to create it from random bytes. 
 - Then, try again to read it. If that also fails then exit with an error message. If it succeeds, continue with those bytes.
 
 That way, homelab instances with work with an only slightly less secure (?) mode by default, but admins can set the env variable to override this or to migrate a database from another instance (if they know what they're doing).
 
Thoughts?

os/server/identifiers/src/jvmMain/kotlin/com/wasmo/identifiers/DbIdentifiers.kt
```diff
@@ -1,11 +1,105 @@
...
+
+/**
+ * A composite value composing salted hash of the password, the salt, and metadata (e.g. hash
+ * algorithm / version used). Because the hashes are salted and peppered, digests of the same
+ * password will in general not be equal.
+ *
+ * Argon2PasswordHasher produces passwordHashes that are salted and peppered
+ */
+@Serializable
+data class PasswordDigest(
```
swankjesse, line 58: For follow-up, I don’t think these types really qualify as identifiers. I would like identifiers to be single-field types `AccountId`, not model types like this. The identifiers package is intended to be the minimum set of common symbols required to compile `api` packages, and these types all seem like api types for a passwords module?

swankjesse: I would like structure modules strictly so the codebase can grow without build performance slowing down

swankjesse: And that is probably my follow-up PR, I have ideas around how to structure this and it might be easier for me to show them than to describe 'em!

os/server/passwords/api/src/jvmMain/kotlin/com/wasmo/passwords/PasswordHasher.kt
```diff
@@ -0,0 +1,54 @@
...
+
+import com.wasmo.identifiers.AccountId
+import com.wasmo.identifiers.AccountPassword
+import com.wasmo.identifiers.HashingPepper
+import com.wasmo.identifiers.HashingSalt
+import com.wasmo.identifiers.PasswordDigest
+import java.security.SecureRandom
+import okio.ByteString.Companion.toByteString
+
+abstract class PasswordHasher(protected val pepper: HashingPepper) {
```
swankjesse, line 11: For follow-up, let’s move this out of `api` because no classes outside of os/server/passwords should use this directly? They’ll use `PasswordStore` ?

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,138 @@
...
+import org.bouncycastle.crypto.PasswordConverter
+import org.bouncycastle.crypto.generators.Argon2BytesGenerator
+import org.bouncycastle.crypto.params.Argon2Parameters
+
+
+/**
+ * Produces salted, peppered [PasswordDigest]s containing metadata about the hashing algorithm
+ * configuration, the raw hash and the salt.
+ *
+ * JSON serialization of example value produced:
```
swankjesse, line 28: I love stuff like this in comments

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,138 @@
...
+    buffer.writeLongLe(accountId.id)
+    return buffer.readByteString()
+  }
+
+  override fun digest(
+    password: AccountPassword,
+    accountId: AccountId,
+  ): PasswordDigest {
+    val salt = saltFactory.create(libraryConfig.saltNumBytes)
+    require(libraryConfig.saltNumBytes == salt.value.size) {
```
swankjesse, line 78: supernit: `check` for checking invariants, `require` for checking parameters. (This is checking an invariant)

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private fun hash(password: AccountPassword, accountId: AccountId, salt: HashingSalt): PasswordHash {
+    val additional: ByteString = additionalBytes(password, accountId)
+    val parameters = libraryConfig.toBouncyCastleParameters(salt, additional)
+    val generator = Argon2BytesGenerator()
+    generator.init(parameters)
+    val passwordHashBytes = ByteArray(libraryConfig.hashNumBytes)
+    val len = generator.generateBytes(
+      /* password= */ password.secretValue.toCharArray(),
```
swankjesse, line 97: Does this accept a char array and not UTF-8 bytes? Ick

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,138 @@
...
+    return PasswordHash(passwordHashBytes.toByteString())
+  }
+
+  override fun verify(
+    password: AccountPassword,
+    accountId: AccountId,
+    passwordDigest: PasswordDigest,
+  ): Boolean {
+    // Guard against backwards compat cases that are not currently implemented.
+    // (As of 2026-08, none are implemented).
```
swankjesse, line 112: 👍🏻

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/Argon2PasswordHasher.kt
```diff
@@ -0,0 +1,138 @@
...
+  }
+}
+
+private fun Argon2LibraryConfig.toBouncyCastleParameters(salt: HashingSalt, additional: ByteString): Argon2Parameters =
+  Argon2Parameters.Builder(type)
+    .withVersion(version)
+    .withMemoryAsKB(memoryKiBytes)
+    .withIterations(iterations)
+    .withParallelism(parallelism)
+    .withCharToByteConverter(PasswordConverter.UTF8)
```
swankjesse, line 135: Ahh, this is where the UTF-8 bytes come back. Phew. You do you, BouncyCastle.

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/PasswordBindings.kt
```diff
@@ -0,0 +1,25 @@
...
+  abstract fun bindPasswordHasher(passwordHasher: Argon2PasswordHasher): PasswordHasher
+
+  companion object {
+    @Provides
+    @SingleIn(OsScope::class)
+    fun providePasswordManager(
+      clock: Clock,
+      passwordHasher: PasswordHasher,
+    ): PasswordStore =
+      RealPasswordStore(clock, passwordHasher)
```
swankjesse, line 22: Can this be a `@Binds` statement by adding `@Inject` to `RealPasswordStore` ?

(I prefer `@Binds` 'cause then if we add more parameters to `RealPasswordStore` we don’t have to update the binding container)

os/server/passwords/real/src/jvmMain/kotlin/com/wasmo/passwords/RealPasswordStore.kt
```diff
@@ -0,0 +1,70 @@
...
+      // It's important that we ignore the result because the verify() used dummy values.
+      val ignored = passwordHasher.verify(providedPassword, accountId, dummyMissingPasswordDigest)
+      isPasswordCorrect
+    }
+  }
+
+  context(sqlTransaction: SqlTransaction)
+  override suspend fun setPassword(
```
swankjesse, line 47: I think the `PasswordStore` is the wrong place for enforcing policy around when you can set a new password. Somewhere else in the code we’re going to let callers set a password from a secret sign-in link, and that call will not have an old password to check.

swankjesse: Not a problem at the moment!

swankjesse: ALSO - perhaps this line of reasoning invalidates my claim that the PasswordStore should be the thing that enforces rate limits? Or perhaps there’s a 3rd thing? Really the thing I’m most concerned with is making it so we do decide-then-execute database interactions instead of a cascade of small business logic decisions.

wasmo-build/ci/Dockerfile
```diff
@@ -16,7 +16,6 @@ FROM eclipse-temurin:25-alpine-3.23
 RUN set -o errexit -o nounset \
     && echo "Installing utilities with apk" \
     && apk add --no-cache \
-      argon2-libs \
```
swankjesse, line 19: Thank you!

(I need to republish our Dockerfiles for this to take effect)

swankjesse (approved):

PR #155, opened 2026-08-30
Declare the ObjectStore API in .wit

swankjesse:


platform/wit/src/wit/object-store/object-store.wit
```diff
@@ -0,0 +1,109 @@
...
+
+  @since(version = 0.1.0)
+  record get-object-request {
+    key: key,
+  }
+
+  @since(version = 0.1.0)
+  record get-object-response {
+    value: list<u8>,
+    etag: option<string>,
```
jingibus, line 34: I don't know etags from ctags, I assume this is something we get locally? We aren't actually making a web request here

platform/wit/src/wit/object-store/object-store.wit
```diff
@@ -0,0 +1,109 @@
...
+  @since(version = 0.1.0)
+  record list-objects-response {
+    entries: list<entry>,
+    next-request: option<list-objects-request>,
+  }
+
+  @since(version = 0.1.0)
+  variant entry {
+    object(entry-object),
+    common-prefix(string),
```
jingibus, line 63: Is there a reference with a similar api that could be linked to? I am not picking up this common prefix delimiter stuff

jingibus (approved):

platform/wit/src/wit/object-store/object-store.wit
swankjesse: This is something we get by making a web request to an S3 service. Or, if we’re doing a local filesystem object store, we generate it by hashing the contents.

platform/wit/src/wit/object-store/object-store.wit
swankjesse: Yep! https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html

(linked below)

PR #162, opened 2026-09-05
Roundtrip all SQL rows through our WIT SQL model

swankjesse:
This is a small first step to using the WIT SQL model everywhere.

Unfortunately the underlying API we're using doesn't give us the type information we need to do this correctly, so this relies on reflection to extract data we don't otherwise have access to. I've structured this reflection hack to fail loudly if the implementation details change in a future release of the Vert.x Postgresql client.


os/server/sql/real/src/jvmMain/kotlin/com/wasmo/sql/VertxRowMetadataHack.kt
```diff
@@ -0,0 +1,37 @@
...
+import io.vertx.sqlclient.Row
+import io.vertx.sqlclient.desc.ColumnDescriptor
+import io.vertx.sqlclient.internal.RowDesc
+import java.lang.reflect.InaccessibleObjectException
+
+/**
+ * Our Vert.x Postgresql client knows the types of each rows' column values, but this data is not
+ * exposed in the public API. Our SQL API needs to encode rows with their type information, so we
+ * use a reflection hack to access a field that's not in the public API.
+ */
```
swankjesse, line 16: https://discord.com/channels/751380286071242794/751398277101912105/1545891452716978277

jingibus: Sigh, this link 404s hard for me. Can't even see what server it is.

swankjesse: It's the Vert.x slack where I'm asking for an API to expose types.
Screenshot_20260910_012139_Discord

PR #170, opened 2026-09-17
Add wkg.toml and tweak witfiles for compatibility

jingibus:


platform/api/src/commonMain/wit/sql/sql.wit
```diff
@@ -29,13 +29,13 @@ interface types {
...
-    f32(f32),
-    f64(f64),
+    boolean(bool),
+    short32(s32),
+    short64(s64),
+    float32(f32),
+    float64(f64),
     datetime(datetime),
-    string(string),
+    sql-string(string),
```
jingibus: All of the types modified here error out in wkg's wit parser as keywords.

That parser is a common library, so it probably makes sense to add that to our validation.

swankjesse:
```suggestion
    %bool(bool),
    %s32(s32),
    %s64(s64),
    %f32(f32),
    %f64(f64),
    datetime(datetime),
    %string(string),
```

swankjesse (approved):

platform/api/src/commonMain/wit/sql/sql.wit
swankjesse: YES

swankjesse:
I think you need to add a @0.1.0 to the UUID reference in build.gradle.kts.

jingibus:
> I think you need to add a @0.1.0 to the UUID reference in build.gradle.kts.

And a lot else besides!

